PR #205, opened 2013-04-02
options: default to core filters

jeffkaufman:
Don't make people write `pagespeed RewriteLevel CoreFilters` in their config.  We should have filters be on by default, like in apache, and people should have to intentionally turn them off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,6 @@ use" below, add this to the top of your configuration:
 In your `nginx.conf`, add to the main or server block:
 
     pagespeed on;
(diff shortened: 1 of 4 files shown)


jeffkaufman:
This was the wrong way to do this, and it's what's causing https://github.com/pagespeed/ngx_pagespeed/issues/1190

`NewRewriteOptions` is used in places where we want a clean `RewriteOptions` that doesn't specify anything, for example, `NewRewriteOptionsForQuery`.  Explicitly setting CoreFilters is the wrong way to make that the default.

Issue #604, opened 2014-01-24
Response HTTP-Header "Date" is always set to Thu, 01 Jan 1970 00:00:00 GMT


Curly060:
ngx_pagespeed always modifies the Response HTTP-Header "Date" to Thu, 01 Jan 1970 00:00:00 GMT. You can verify this easily with your demonstration site:
- Pagespeed on: http://ngxpagespeed.com/ngx_pagespeed_example/add_instrumentation.html?ModPagespeed=on&ModPagespeedFilters=add_instrumentation
- Pagespeed off: http://ngxpagespeed.com/ngx_pagespeed_example/add_instrumentation.html?ModPagespeed=off&ModPagespeedFilters=add_instrumentation

This breaks behaviour with cookies when the cooke max age is set to 0m because the cookies age will then be somewhere in year 2057 instead of in the past (which would delete the cookie).

Is there any workaroung for this or some fix?

Issue #577, opened 2013-12-14
Wrong date HTTP header after HTML rewrite


mathieubruno:
The **Date HTTP Header** seems to be wrong because it seems to be copied in ngx_pagespeed.cc. Shouldn't it be update with the server date (with **gmtime** and **strftime "%a, %d %b %Y %H:%M:%S GMT"** for example) ?

(Redbot.org told me the date was wrong for the pagespeed resources)

PR #605, opened 2014-01-24
Date header: use current date when we don't get one handed over

oschaaf:
When the content generator does not supply us with a date header,
we need to create one ourselves and set it to the current date.

Fixes:
https://github.com/pagespeed/ngx_pagespeed/issues/604 (duplicate)
https://github.com/pagespeed/ngx_pagespeed/issues/577

### DIFF
--- a/test/pagespeed_test.conf.template
+++ b/test/pagespeed_test.conf.template
@@ -616,6 +616,13 @@ http {
   }
 
   server {
+    listen @@SECONDARY_PORT@@;
+    server_name date.example.com;
+    pagespeed FileCachePath "@@FILE_CACHE@@";
+    add_header "Date" "Date: Fri, 16 Oct 2009 23:05:07 GMT";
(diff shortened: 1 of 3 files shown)


test/pagespeed_test.conf.template
```diff
@@ -616,6 +616,13 @@ http {
   }
 
   server {
+    listen @@SECONDARY_PORT@@;
+    server_name date.example.com;
+    pagespeed FileCachePath "@@FILE_CACHE@@";
+    add_header "Date" "Date: Fri, 16 Oct 2009 23:05:07 GMT";
```
oschaaf: Note: This `add_header` line looks wrong, which is already fixed on trunk-tracking via https://github.com/pagespeed/ngx_pagespeed/commit/a8141eadd3e22a66fcbc5dacecf97b9669c3a5ef

PR #702, opened 2014-05-13
content-length: when we don't preserve headers, clear the CL header

oschaaf:
I haven't been able to reproduce a problem with this, looking at the code it seems to me that we do need to do this.


oschaaf:
This probably is outdated - closing

Issue #770, opened 2014-08-07
@psol-etag breaks AWS Elastic Load Balancing


neofreko:
AWS Load Balancer cannot handle '@' in the header field. I can confirm removing the '@' fixes it but I don't know whether it will affect any pagespeed functionalities.

If the '@' part is intentional, I hope to get some sort of option to modify it.

Below is a sample of valid nginx response but made ELB throws 502 bad Gateway
```
[ root@2344a5b00ff6:/etc/nginx ]$ HEAD http://localhost/css/A.common.css.pagespeed.cf.tEbyNOHJUq.css
200 OK
Cache-Control: max-age=315360000
Cache-Control: public, must-revalidate, proxy-revalidate
Connection: close
Date: Thu, 07 Aug 2014 13:11:09 GMT
Pragma: public
Pragma: public
Accept-Ranges: bytes
Server: Tengine/2.0.3
Content-Type: text/css
Expires: Thu, 31 Dec 2037 23:55:55 GMT
Last-Modified: Thu, 07 Aug 2014 13:11:09 GMT
@Psol-Etag: W/"0"
Client-Date: Thu, 07 Aug 2014 13:15:47 GMT
Client-Peer: 127.0.0.1:80
Client-Response-Num: 1
X-Page-Speed: 1.8.31.4-4009
```

PR #816, opened 2014-09-26
gzip-vs-etag: Don't rename the ETag when gzip isn't ./configured

oschaaf:
We should not rename the ETag header to work around the gzip module
clearing it, when the gzip module actually isn't built. In that case
we will fail to inject the header filter that renames the header back
to 'ETag' before it gets send out.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/770


jeffkaufman:
Is this still good to go in?  LGTM

PR #834, opened 2014-10-20
NgxConnection: single event mechanism for NgxFetch and NgxBaseFetch

oschaaf:
Abstract the pipe communication into NgxEventConnection, for reuse
by NgxBaseFetch and NgxUrlAsyncFetcher.

Based on Chai's earlier work, but with a few fixes discovered
while working on this and SPDY module compatibility
- Uses less file descriptors, I expect this to be faster but need
  measurement is needed to back that.
- Fixed NgxUrlAsyncFetcher actually shutting down its fetchers.
- Fixes a bug where we wouldn't clean idle pooled NgxConnections.
- Fixes a bug for requests that are unlucky enough to be finalized mid-IPRO lookup.
- Makes us use ngx_handle_read_event/ngx_del_event

Note: this needs a few more manual tests,  and I also want to ensure 
this uses ngx_handle_read_event/ngx_del_event correctly. That aside,
I think this is ready for review.

### DIFF
--- a/config
+++ b/config
@@ -169,6 +169,7 @@ if [ $ngx_found = yes ]; then
     $ps_src/log_message_handler.h \
     $ps_src/ngx_base_fetch.h \
     $ps_src/ngx_caching_headers.h \
+    $ps_src/ngx_event_connection.h \
     $ps_src/ngx_fetch.h \
     $ps_src/ngx_gzip_setter.h \
     $ps_src/ngx_list_iterator.h \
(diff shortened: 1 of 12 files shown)


jeffkaufman:
This change is causing frequent valgrind errors on my machine, even without parallel tests.  My machine is heavily loaded because I'm running some fuzzers, but with this change in I get valgrind errors on 9 of 20 test runs and with it rolled back I got 0 of 20.  I've sent the valgrind error logs to @oschaaf .

I'm going to roll this back for now, we can roll forward when we figure out the leak.

PR #869, opened 2015-01-06
mod_pagespeed moved the location of the system directory; update our includes

jeffkaufman:


src/ngx_rewrite_driver_factory.h
```diff
@@ -28,13 +28,9 @@ extern "C" {
 
 #include <set>
 
-#include "net/instaweb/system/public/system_rewrite_driver_factory.h"
 #include "pagespeed/kernel/base/md5_hasher.h"
 #include "pagespeed/kernel/base/scoped_ptr.h"
-
-// TODO(oschaaf): We should reparent ApacheRewriteDriverFactory and
-// NgxRewriteDriverFactory to a new class OriginRewriteDriverFactory and factor
-// out as much as possible.
```
jeffkaufman: This TODO is obsolete; we already did that.  (But we called it SystemRDF instead of OriginRDF.)

oschaaf:
LGTM

PR #872, opened 2015-01-09
Revert "NgxConnection: single event mechanism

jeffkaufman:
This reverts commit 5dfe42f3d34a4f43dc90d73e8bfa79116610a7a7 (#834).

With #834 I got memory errors with valgrind on 9 of 20 runs; without it I got 0 of 20.  When we fix up the memory issue we can re-apply.


oschaaf:
LGTM

PR #876, opened 2015-01-16
Oschaaf single fd fixed

oschaaf:
Re-applies the single fd patch, plus a commit  which fixes the `NgxBaseFetch` leak that occurred in the system tests at the point it reloaded configuration.


jeffkaufman:
LGTM

Issue #864, opened 2014-12-25
ngx_pagespeed overrides Server header


akuryan:
My nginx -V output:
nginx version: nginx/1.6.2
built by gcc 4.4.7 20120313 (Red Hat 4.4.7-11) (GCC)
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-http_ssl_module --with-http_realip_module --with-http_addition_module --with-http_sub_module --with-http_dav_module --with-http_flv_module --with-http_mp4_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_random_index_module --with-http_secure_link_module --with-http_stub_status_module --with-mail --with-mail_ssl_module --with-http_spdy_module --add-module=/tmp/tmp1ra9oc/rpmbuild//BUILD/nginx-1.6.2/headers-more-nginx-module-0.25 --add-module=/tmp/tmp1ra9oc/rpmbuild//BUILD/nginx-1.6.2/ngx_pagespeed-1.8.31.4-beta --with-file-aio --with-ipv6 --with-cc-opt='-O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic'

If pagespeed is turned on, it sets Server header to 'nginx' for all dynamic resources.
It does not matter if I am trying to modify this header by means of changing nginx sources or by header-more module, as suggested in accepted answer here:
http://stackoverflow.com/questions/246227/how-do-you-change-the-server-header-returned-by-nginx

Is it possible to modify Server header to custom one with pagespeed turned on?

akuryan:
@oschaaf Just wanted to inform you that this patch made it's way through and finally are playing on our production

oschaaf:
[...]

PR #879, opened 2015-01-16
server-header: improve handling of server header

oschaaf:
- Fixes an issue in the html flow that would prevent overriding the
value of the 'Server' response-header.
- Add tests that ensure we emit a single and correct server header
in all flows when not overriding it.
- Add tests that ensure overriding the 'Server' response header
works. The resource and IPRO flow are added to the expected failures
as those are not working yet (and will be adressed in a follow-up).

Fixed https://github.com/pagespeed/ngx_pagespeed/issues/864 (html flow)

### DIFF
--- a/test/nginx_system_test.sh
+++ b/test/nginx_system_test.sh
@@ -274,6 +274,8 @@ STATISTICS_URL=http://$PRIMARY_HOSTNAME/ngx_pagespeed_statistics
 
 # An expected failure can be indicated like: "~In-place resource optimization~"
 PAGESPEED_EXPECTED_FAILURES="
+~Override server header in resource flow.~
+~Override server header in IPRO flow.~
 "
 
 # Some tests are flakey under valgrind. For now, add them to the expected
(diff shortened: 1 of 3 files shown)


test/nginx_system_test.sh
```diff
@@ -1101,6 +1103,47 @@ start_test Base config has purging disabled.  Check error message syntax.
 OUT=$($WGET_DUMP "$HOSTNAME/pagespeed_admin/cache?purge=*")
 check_from "$OUT" fgrep -q "pagespeed EnableCachePurge on;"
 
+start_test Default server header in html flow.
+URL="http://$SECONDARY_HOSTNAME/mod_pagespeed_example/"
+HEADERS="--header=Host:headers.example.com"
+OUT=$($WGET_DUMP -O /dev/null -S $HEADERS $URL 2>&1)
```
jeffkaufman: Could you write this test with `http_proxy=` instead?  I know we have some of both and they're technically equivalent, but we've been trying to standardize some.

So:

```
URL=http://headers.example.com/mod_pagespeed_example/
OUT=$(http_proxy=$SECONDARY_HOSTNAME
    $WGET_DUMP -O /dev/null -S $URL 2>&1)
```

jeffkaufman:
LGTM

test/nginx_system_test.sh
oschaaf: Done

Issue #612, opened 2014-01-31
Ignoring other modules manipulating the headers


michelkaeser:
I'm not sure weither this is a bug or related to ngx_pagespeed, but I'm opening this issue as a result of the discussion over here: https://github.com/agentzh/headers-more-nginx-module/issues/21#issuecomment-33710459

I've seen that PageSpeed forces the existence of the X-PageSpeed header - but may it be that the current implementation of this "assurance" is somehow also manipulating the other headers?
For example unsetting the 'Server' header doesn't work when using PageSpeed.

Thanks!

oschaaf:
https://github.com/pagespeed/ngx_pagespeed/pull/906 just landed on trunk-tracking, and adds a switch to position ngx_pagespeed at a different position in the module order. This allows header modules to fully control the output.
Anyone willing to give it a spin?

PR #880, opened 2015-01-19
posted-requests: run posted requests when appropriate

oschaaf:
Request termination code relies on posted requests being run.
One of the things this fixes is that we won't hang when configured
to run before the SSI module.

Avoids trouble when changing the module order, which helps
https://github.com/pagespeed/ngx_pagespeed/issues/612


jeffkaufman:
LGTM

PR #885, opened 2015-01-23
Release base fetch early, eliminate write_pending and fetch_done

oschaaf:
Get rid of write_pending & fetch_done flags on the request context.
Assume that having ctx->base_fetch set means we have outstanding
work and thus we are not done yet, for simplicity.
This also means we'll release the base fetch earlier which seems
like a good idea to me.


jeffkaufman:
LGTM

PR #899, opened 2015-02-05
trunk-tracking-shutdown: improve shutdown code

oschaaf:
- Add a test to make sure all logged output looks sane, whitelisting current errors/warnings.
- Stop our nginx test instances after we are done testing.
- Add tests for shutting down and reloading configuration under high load (depends on ab).
- Reduce the number of keepalive requests in the keepalive tests to speed up test runs.
- Fix exiting with open file descriptors, fix cleanup in nginx's cache manager/loader processes
- Attempt to finish up queued up NgxBaseFetches/requests on shutdown/reload with a timeout
- Under valgrind the blocking rewrite started failing after adding a test for reloading configuration under high load. 

This is an incremental improvement, the shutdown still needs further improvement.

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -85,6 +85,8 @@ extern ngx_module_t ngx_pagespeed;
 // Needed for SystemRewriteDriverFactory to use shared memory.
 #define PAGESPEED_SUPPORT_POSIX_SHARED_MEM
 
+net_instaweb::NgxRewriteDriverFactory* active_driver_factory = NULL;
+
 namespace net_instaweb {
 
 const char* kInternalEtagName = "@psol-etag";
@@ -880,11 +882,12 @@ void ps_cleanup_srv_conf(void* data) {
   // to be shut down when we destroy any proxy_fetch_factories. This
   // will prevent any queued callbacks to destroyed proxy fetch factories
   // from being executed
-
   if (!factory_deleted && cfg_s->server_context != NULL) {
+    if (active_driver_factory == cfg_s->server_context->factory()) {
+      active_driver_factory = NULL;
+    }
     delete cfg_s->server_context->factory();
     factory_deleted = true;
-    NgxBaseFetch::Terminate();
   }
   if (cfg_s->proxy_fetch_factory != NULL) {
     delete cfg_s->proxy_fetch_factory;
@@ -931,12 +934,20 @@ void ps_set_conf_cleanup_handler(
 }
 
 void terminate_process_context() {
+  if (active_driver_factory != NULL) {
+    // If we got here, that means we are in the cache loader/manager
+    // or did not get a chance to cleanup otherwise.
+    delete active_driver_factory;
+    active_driver_factory = NULL;
+    NgxBaseFetch::Terminate();
+  }
   delete process_context;
   process_context = NULL;
 }
 
 void* ps_create_main_conf(ngx_conf_t* cf) {
   if (!process_context_cleanup_hooked) {
+    SystemRewriteDriverFactory::InitApr();
     atexit(terminate_process_context);
     process_context_cleanup_hooked = true;
   }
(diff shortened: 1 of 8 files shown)


src/ngx_pagespeed.cc
```diff
@@ -2967,10 +2979,18 @@ ngx_int_t ps_init_module(ngx_cycle_t* cycle) {
...
+    active_driver_factory = NULL;
   }
   return NGX_OK;
 }
 
+void ps_exit_child_process(ngx_cycle_t* cycle) {
+  ps_main_conf_t* cfg_m = static_cast<ps_main_conf_t*>(
+      ngx_http_cycle_get_module_main_conf(cycle, ngx_pagespeed));
+  NgxBaseFetch::Terminate();
+  cfg_m->driver_factory->ShutDown();
```
oschaaf: @jeffkaufman 
I do think we need to remove the DCHECK in SystemCaches::ShutDown(): I need to explicitly call ShutDown() on the driver factory here and destruct it at another point in time (pool cleanup handler). When NgxBaseFetch::Terminate() isn't able to round up within its timeout period, it should shutdown the driver factory as soon as possible.

jeffkaufman: As discussed with ac4b99d, this `DCHECK` can stay.

test/nginx_system_test.sh
```diff
@@ -523,11 +528,24 @@ check test $(scrape_stat image_rewrite_total_original_bytes) -ge 10000
 # happens both before and after.
 start_test "Reload config"
 
+AB_PID="0"
+# Fire up some heavy load if ab is available to test a stressed shutdown
+if hash ab 2>/dev/null; then
+    ab -n 10000 -c 100 -k http://$PRIMARY_HOSTNAME/ &>/dev/null & AB_PID=$!
+    # Sleep to allow some queueing up of requests
+    sleep 2
+fi
```
jeffkaufman: If `ab` isn't present, could we print something about being sad we can't run this test?

test/nginx_system_test.sh
```diff
@@ -1095,8 +1113,8 @@ check_not_from "$OUT" fgrep -qi 'addInstrumentationInit'
 
 if [ "$NATIVE_FETCHER" != "on" ]; then
   start_test Test that we can rewrite an HTTPS resource.
-  fetch_until $TEST_ROOT/https_fetch/https_fetch.html \
-    'grep -c /https_gstatic_dot_com/1.gif.pagespeed.ce' 1
+  # fetch_until $TEST_ROOT/https_fetch/https_fetch.html \
+  #  'grep -c /https_gstatic_dot_com/1.gif.pagespeed.ce' 1
```
jeffkaufman: ?

test/nginx_system_test.sh
```diff
@@ -1153,19 +1171,81 @@ check_from "$OUT" fgrep -qi '404'
...
 
+start_test Shutting down.
+
+AB_PID="0"
+# Fire up some heavy load if ab is available to test a stressed shutdown
+if hash ab 2>/dev/null; then
+    ab -n 10000 -c 100 http://$PRIMARY_HOSTNAME/ &>/dev/null & AB_PID=$!
+    # Sleep to allow some queueing up of requests
+    sleep 2
+fi
```
jeffkaufman: some warning should go here

jeffkaufman: in fact, can we pull this out into a helper function?

test/nginx_system_test.sh
```diff
@@ -523,11 +528,24 @@ check test $(scrape_stat image_rewrite_total_original_bytes) -ge 10000
 # happens both before and after.
 start_test "Reload config"
 
+AB_PID="0"
+# Fire up some heavy load if ab is available to test a stressed shutdown
```
jeffkaufman: stressed shutdown -> stressed config reload

jeffkaufman:
LGTM

test/nginx_system_test.sh
oschaaf: Whoops, left in the https fetching test commented by accident, which I had to do because my dev box is a bit messed up and can't validate the certificates.

PR #901, opened 2015-02-05
Support for pushState(); and popState(); ajax loaded sites

Gonzo2028:
My newest project is using pushState(), and popState(); and it loads only the changing middle part of the page instead of reloading it entirely. This is not really supported by pagespeed.

Watch my example.
https://mysexyescort.de/

Plus why its it impossible to send expire headers to clients when running pagespeed?
I 'd like to cache their get requests for 2 minutes or so.


jeffkaufman:
This pull request seems to move random commits from `master` to an already released branch.  Did you create it by mistake?

Gonzo2028:
Hey,

hm no, but maybe i did it at the wrong place?
Regards

Am 05.02.2015 um 17:43 schrieb Jeff Kaufman:

PR #905, opened 2015-02-10
Port remote configuration option to nginx.

crowell:
Ports config option and tests to nginx.

depends on the addition to psol first.

still tracking down crash when the remote server is on the same nginx instance as the server fetching the config. (this issue doesn't show up on apache)

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -203,6 +203,145 @@ ngx_int_t string_piece_to_buffer_chain(
   return NGX_OK;
 }
 
+// Get the context for this request.  ps_connection_read_handler should already
+// have been called to create it.
+ps_request_ctx_t* ps_get_request_context(ngx_http_request_t* r) {
+  return static_cast<ps_request_ctx_t*>(
+      ngx_http_get_module_ctx(r, ngx_pagespeed));
+}
+
+// Tell nginx whether we have network activity we're waiting for so that it sets
+// a write handler.  See src/http/ngx_http_request.c:2083.
+void ps_set_buffered(ngx_http_request_t* r, bool on) {
+  if (on) {
+    r->buffered |= NGX_HTTP_PAGESPEED_BUFFERED;
+  } else {
+    r->buffered &= ~NGX_HTTP_PAGESPEED_BUFFERED;
+  }
+}
+
+namespace ps_base_fetch {
+
+ngx_http_output_header_filter_pt ngx_http_next_header_filter;
+ngx_http_output_body_filter_pt ngx_http_next_body_filter;
+
+ngx_int_t ps_base_fetch_filter(ngx_http_request_t* r, ngx_chain_t* in) {
+  ps_request_ctx_t* ctx = ps_get_request_context(r);
+
+  if (ctx == NULL || ctx->base_fetch == NULL) {
+    return ngx_http_next_body_filter(r, in);
+  }
+
+  ngx_log_debug1(NGX_LOG_DEBUG_HTTP, r->connection->log, 0,
+                 "http pagespeed write filter \"%V\"", &r->uri);
+
+  // send response body
+  if (in || ctx->write_pending) {
+    ngx_int_t rc = ngx_http_next_body_filter(r, in);
+    ctx->write_pending = (rc == NGX_AGAIN);
+    if (rc == NGX_OK && !ctx->fetch_done) {
+      return NGX_AGAIN;
+    }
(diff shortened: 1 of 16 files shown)


crowell:
the crash mentioned in the first comment was due to `worker_processes` being set to 1.

I talked to @jeffkaufman offline and am bumping the value to 2 for this test so that the test can be self-contained and not rely on a secondary server.

src/ngx_pagespeed.cc
```diff
@@ -1265,6 +1265,22 @@ ps_loc_conf_t* ps_get_loc_config(ngx_http_request_t* r) {
...
+  if (!cfg_s || !cfg_s->server_context ||
+      !cfg_s->server_context->global_options()) {
+    return NULL;
+  }
+  if (!cfg_s->server_context->global_options()
+          ->remote_configuration_url()
+          .empty()) {
+    RewriteOptions* remote_options =
+        cfg_s->server_context->global_options()->Clone();
+    cfg_s->server_context->GetRemoteOptions(remote_options);
```
jeffkaufman: Can you comment that this is a blocking fetch?  Things like that are very unusual here.

test/nginx_system_test.sh
```diff
@@ -252,7 +252,7 @@ fi
 
 # check_stat in system_test_helpers.sh needs to know whether statstistics are
 # enabled, which is always the case for ngx_pagespeed.
-statistics_enabled=1
+statistics_enabled=0
```
jeffkaufman: This isn't a good fix for the statistics issue.  As far as I know only the cache flushing test has a problem, but here we disable all stats testing in ngx_pagespeed.

test/pagespeed_test.conf.template
```diff
@@ -1058,6 +1059,59 @@ http {
...
   }
 
+  # Start remote config servers.
+  server {
+    listen @@SECONDARY_PORT@@;
+    listen [::]:@@SECONDARY_PORT@@;
+    server_name remote-config.example.com;
+    pagespeed FileCachePath "@@SECONDARY_CACHE@@";
+    pagespeed RemoteConfigurationUrl "http://localhost:@@PRIMARY_PORT@@/mod_pagespeed_test/remote_config/remote.cfg?PageSpeed=off";
+    #pagespeed RemoteConfigurationUrl "http://raxcity.com/remote.cfg";
```
jeffkaufman: remove before submitting

test/pagespeed_test.conf.template
```diff
@@ -1058,6 +1059,59 @@ http {
...
+  }
+  server {
+    listen @@SECONDARY_PORT@@;
+    listen [::]:@@SECONDARY_PORT@@;
+    server_name remote-config-partially-invalid.example.com;
+    pagespeed FileCachePath "@@SECONDARY_CACHE@@";
+    pagespeed RemoteConfigurationUrl "http://localhost:@@PRIMARY_PORT@@/mod_pagespeed_test/remote_config/remote.cfg.partially.invalid";
```
jeffkaufman: shouldn't this one and the ones below all have `?PageSpeed=off` ?

jeffkaufman: Or is the exception in `location /mod_pagespeed_test/remote_config/` enough, in which case we can remove it from L1068 above?

crowell: moved tests to use nc.

test/pagespeed_test.conf.template
```diff
@@ -1078,6 +1078,57 @@ http {
     pagespeed DisableFilters remove_comments,add_instrumentation;
   }
 
+  # Start remote config servers.
+  server {
+    listen @@SECONDARY_PORT@@;
+    listen [::]:@@SECONDARY_PORT@@;
+    server_name remote-config.example.com;
+    pagespeed FileCachePath "@@SECONDARY_CACHE@@";
+    pagespeed RemoteConfigurationUrl "http://127.0.0.1:9991/remote.cfg";
```
jeffkaufman: In the apache test you made these ports be substituted variables; want to do that here too?

crowell:
$RCPORTN added.

jeffkaufman:
LGTM.

Could you squash the commit history and force push?  There are a lot of commits in there that don't need to end up in trunk.

(Don't merge until the PSOL side of the change is in, though.)

crowell:
ugh, broke it squashing, redoing in a minute

crowell:
deleting this one, see #925 (same code pre breaking squash)

PR #906, opened 2015-02-12
module-position: optional switch to install at HTTP_AUX_FILTER_MODULES

oschaaf:
This change adds a flag to optionally position the module at HTTP_AUX_FILTER_MODULES
instead of just in front of the gzip module. Doing so allows header module(s) to
fully manipulate any response headers before they are send to the user-agent.
With https://github.com/pagespeed/ngx_pagespeed/pull/880, SSI no longer hangs,
and other modules using subrequests should not do so either. SSI subrequests are
not optimized when ngx_pagespeed runs at HTTP_AUX_FILTER_MODULES, so the test for
it is added to the expected failures when the tests are run for this module position.

To build with the module respositioned:
  POSITION_AUX=true ./configure ......
The same variable can be set before running the system tests, in which case it will
perform additional header tests and add the SSI test to the expected failures.

### DIFF
--- a/config
+++ b/config
@@ -17,6 +17,8 @@
 #   PSOL_BINARY: absolute path to pagespeed_automatic.a
 
 mod_pagespeed_dir="${MOD_PAGESPEED_DIR:-unset}"
+position_aux="${POSITION_AUX:-unset}"
+
 if [ "$mod_pagespeed_dir" = "unset" ] ; then
   mod_pagespeed_dir="$ngx_addon_dir/psol/include"
   build_from_source=false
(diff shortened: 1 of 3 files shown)


jeffkaufman:
What's the reason for making it configurable?  Why not just move us there always?

oschaaf:
@jeffkaufman I'm pretty confident things work fine at the new module position, but with this approach I was hoping that perhaps someone who commented at https://github.com/pagespeed/ngx_pagespeed/issues/612 could confirm some production mileage before flipping the switch. What do you think?

oschaaf:
@jeffkaufman Also, it would be nice to get SSI subrequests to be optimized (though I'm not sure how much this is actually used)

jeffkaufman:
LGTM

PR #907, opened 2015-02-12
configuration-reload-test: wait for the new worker to get ready

oschaaf:
After sending a SIGHUP, wait for the new worker process with the
new configuration to handle further requests, or else the tests
running after that might get confused when the new worker becomes
active with fresh shared mem statistics and shared mem cache.

### DIFF
--- a/test/nginx_system_test.sh
+++ b/test/nginx_system_test.sh
@@ -302,6 +302,9 @@ fi
 #
 # TODO(sligicki): When the prioritize critical css race condition is fixed, the
 # two prioritize_critical_css tests no longer need to be listed here.
+# TODO(oschaaf): Now that we wait after we send a SIGHUP for the new worker 
+# process to handle requests, check if we can remove more from the expected 
+# failures here under valgrind.
 if $USE_VALGRIND; then
     PAGESPEED_EXPECTED_FAILURES+="
 ~combine_css Maximum size of combined CSS.~
@@ -310,7 +313,6 @@ if $USE_VALGRIND; then
 ~IPRO flow uses cache as expected.~
 ~IPRO flow doesn't copy uncacheable resources multiple times.~
 ~inline_unauthorized_resources allows unauthorized css selectors~
-~Blocking rewrite enabled.~
 "
 fi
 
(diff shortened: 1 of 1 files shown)


test/nginx_system_test.sh
```diff
@@ -551,6 +553,11 @@ fire_ab_load
 check wget $EXAMPLE_ROOT/styles/W.rewrite_css_images.css.pagespeed.cf.Hash.css \
   -O /dev/null
 check_simple "$NGINX_EXECUTABLE" -s reload -c "$PAGESPEED_CONF"
+# Wait for the new worker process with the new configuration to get ready, or
+# else the sudden reset of the shared mem statistics/cache might catch upcoming
+# tests unaware.
+fetch_until $STATISTICS_URL \
+  'grep -c image_rewrite_total_original_bytes:[[:space:]]*0' 1
```
jeffkaufman: Don't we have a scrape_stat function for this?

jeffkaufman:
This looks great!  I think this might fix the intermittent cache_flush test error we've seen.

(Specifically, we'd see the cache flush properly, but the stat wouldn't get bumped.  That could be explained by the fresh stats coming online at exactly the wrong time, so that the flush happened on the old stats and the query to the stats url happened on the new stats.  This was especially likely because that test has some sleeps.)

test/nginx_system_test.sh
oschaaf: Re: "Don't we have a scrape_stat function for this?"
Yes - updated. PTAL?

oschaaf:
@jeffkaufman "I think this might fix the intermittent cache_flush test error we've seen."
That sounds like something this could very well fix. I've only seen that flake one or two times in the past weeks, so it's hard to be sure.

jeffkaufman:
LGTM!

oschaaf:
@jeffkaufman Ugh. With this merged, the sanity check for `error.log` starts failing when running the tests in a loop under valgrind. Inspecting `error.log`, it looks like somehow the old worker process isn't able to finish up all queued requests in the timeout period. However, it does not look like that happens because it's short on time. Some requests from `ab` just seem to stall, so I think I need to look at if there's an edge case that I missed.

PR #908, opened 2015-02-13
shutdown: fix system test failure in error.log sanity check

oschaaf:
With this change tests should finish with sane logging for the current test workload, 
which means all pending requests and `NgxBaseFetch` instances are handled.
We might also want to add more tests for the shutdown and reload where:
  - very slow to respond pages are proxied.
  - very long running chunky html pages are proxies
  - with pages that reference very slow resources to stress the shutdown of the fetchers

This change also has some minor cleanup


jeffkaufman:
LGTM

Issue #913, opened 2015-02-15
Regression: Pagespeed sets invalid/incorrect Date header


mqudsi:
This issue is possibly related to #577 and #604/#605 

With ngx_pagespeed, the Date header is being output incorrectly.

```
curl -I http://squdsi.com/
HTTP/1.1 200 OK
Server: nginx/1.7.9
Content-Type: text/html; charset=utf-8
Connection: keep-alive
Vary: Accept-Encoding
Date: Wed, 28 Jan 1970 10:26:51 GMT
X-Page-Speed: 1.9.32.3-4448
Cache-Control: max-age=0, no-cache
```

vs with ngx_pagespeed disabled:
```
curl -I "http://squdsi.com/?ModPagespeed=off&ModPagespeedFilters=add_instrumentation"
HTTP/1.1 200 OK
Server: nginx/1.7.9
Date: Sun, 15 Feb 2015 12:51:04 GMT
Content-Type: text/html
Content-Length: 2577
Last-Modified: Sat, 26 May 2012 15:22:52 GMT
Connection: keep-alive
Vary: Accept-Encoding
ETag: "4fc0f54c-a11"
Vary: Accept-Encoding
Accept-Ranges: bytes
```

I've had to turn off ngx_pagespeed for the moment until I figure out why it's segfaulting with every other request (I'll report if it's not my mistake), but I'll try and provide as much debug info as I can.

oschaaf:
@mqudsi It would be great of you could recompile `--with-debug`, and add `pagespeed InstallCrashHandler on;` to `nginx.conf`. This probably would get a backtrace in `error.log` for the segfault, which in turn hopefully helps a lot figuring out why that happens.

oschaaf:
@mqudsi About the wrong date header: I'll look into it.

mqudsi:
This is probably not going to be of much help.

```
nginx: worker process() [0x8075f0d]
nginx: worker process() [0x806c17e]
nginx: worker process() [0x8073cae]
nginx: worker process() [0x80724f8]
nginx: worker process() [0x8074c44]
nginx: worker process() [0x8054d7e]
/lib/i386-linux-gnu/libc.so.6(__libc_start_main+0xf3) [0x3ad4d3]
nginx: worker process() [0x8053aa1]
```

That's with `--with-debug`. I was planning on enabling core dumps and running under gdb, unless you have a better way?

oschaaf:
[...]

PR #916, opened 2015-02-17
date-header: fix 32 bits issue

oschaaf:
Should fix https://github.com/pagespeed/ngx_pagespeed/issues/913


jeffkaufman:
Yes, it does looks like ngx_current_msec is just the least significant bits on 32-bit: http://osdir.com/ml/nginx-development/2010-05/msg00019.html

Another option here would have been the `(uint64_t) ngx_cached_time->sec * 1000 + ngx_cached_time->msec` proposed in that thread, but I'm happy with `PosixTimer` here.

jeffkaufman:
LGTM

PR #917, opened 2015-02-17
native-fetcher: bump dns timeout to deflake running the tests

oschaaf:
Nginx sometimes receives an expected response from 8.8.8.8 when
resolving host names, and the subsequent retry doesn't finish up
in 10 seconds. Bumping the timeout helps, but does not fully
explain why nginx is sometimes receiving an unexpected response.
I'll create a separate issue for investigating that.


crowell:
lgtm

Issue #918, opened 2015-02-17
native-fetcher: investigate "unexpected response" messages in error.log 


oschaaf:
Sometimes nginx's resolver used by the native fetcher complains about receiving unexpected responses from Google's 8.8.8.8 dns server. It would be good to know why that happens to ensure all is well.

https://github.com/pagespeed/ngx_pagespeed/pull/917 adds a line to allow this error message in `error.log` after running the system tests.

zarza:
Log is showing issues with this. I changed dns server to 4.2.2.1 and no additional errors... I guess this should be addressed soon.

PR #922, opened 2015-02-19
native-fetcher: fix a few bugs

oschaaf:
While starting to look into https://github.com/pagespeed/ngx_pagespeed/issues/918,
I noticed a few things that looked off. So:

- Add handling for header only responses
- Fixup i/o event handling
- Fixup NgxFetch's state transitions while handling i/o

Should help stabilizing (valgrind) test runs some more.
Native-fetcher specific tests will be added as a follow-up to this, after sorting out #918 .

### DIFF
--- a/src/ngx_fetch.cc
+++ b/src/ngx_fetch.cc
@@ -131,7 +131,7 @@ NgxConnection* NgxConnection::Connect(ngx_peer_connection_t* pc,
         connection_pool.Remove(nc);
 
         ngx_log_error(NGX_LOG_DEBUG, pc->log, 0,
-                      "NgxFetch: re-using connection %p (pool size: %l)\n",
+                      "NgxFetch: re-using connection %p (pool size: %l)",
                       nc, connection_pool.size());
         return nc;
       }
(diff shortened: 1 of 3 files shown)


src/ngx_fetch.cc
```diff
@@ -735,31 +735,34 @@ void NgxFetch::ConnectionWriteHandler(ngx_event_t* wev) {
...
-        fetch->CallbackDone(false);
-      }
-      // Timer set in Init() is still in effect.
-      return;
+      break;
     } else {
-      c->error = 1;
-      fetch->CallbackDone(false);
-      return;
+      ok = false;
```
jeffkaufman: is this important enough to log anything over?

oschaaf: @jeffkaufman Might be, currently we'll log debug messages at a higher level NgxFetch::CallbackDone
I also *think* that nginx will log something when we end up here, though it might be hardish to associate that message with this NgxFetch instance. So, should I just add a warning message?

jeffkaufman: I think a warning would be good.

jeffkaufman:
LGTM

PR #925, opened 2015-02-23
Port remote configuration option to nginx.

crowell:
Deleting other one that I broke squashing.

PR #931, opened 2015-03-05
native-fetcher: fixes

oschaaf:
- Native fetching broke in https://github.com/pagespeed/ngx_pagespeed/pull/925
  Make changes so we can do early-stage fetches in the startup process
  to fix that.
- Properly cancel out any pending fetches upon ShutDown()
- Decline fetching after ShutDown(), decline fetching when Init()
  failed.


crowell:
LGTM, fixes the crash from #925.

PR #932, opened 2015-03-05
rm "unexpected response" grep from system test

crowell:
remove unneeded "unexpected response" inverse grep from system test that was readded in #925


oschaaf:
LGTM

PR #934, opened 2015-03-13
Moved `automatic/` into `pagespeed/` and got rid of the `public/` hierarchy.

jeffkaufman:


jeffkaufman:
The svn changes for this are already submitted (my mistake) so ngx_pagespeed won't build without this patch applied.

test/nginx_system_test.sh
```diff
@@ -289,7 +289,6 @@ SERVER_NAME=nginx
 
 # run generic system tests
 SYSTEM_TEST_FILE="$MOD_PAGESPEED_DIR/src/pagespeed/system/system_test.sh"
-AUTOMATIC_SYSTEM_TEST="$MOD_PAGESPEED_DIR/src/net/instaweb/automatic/system_test.sh"
```
morlovich: What's the deal with this?

jeffkaufman: We used to need to tell `system/system_test.sh` how to find `automatic/system_test.sh`, but now that `automatic/` and `/system` are in sibling directories `system/system_test.sh` can do it itself and ignores the `AUTOMATIC_SYSTEM_TEST` variable.

crowell:
LGTM

PR #936, opened 2015-03-16
shutdown-and-flushing: Improve shutdown, support/test FlushHtml

oschaaf:
shutdown-and-flushing: Improve shutdown, support/test FlushHtml

```
- Improve shutting down both gracefully and quickly
- Set the flush flags on buffers when FlushHtml is on for optimized
  html responses
- Add tests for these features
- Don't call proxy_fetch->Flush() each time our body filter is called,
  because that would only make sense if we'd also set buf->flush.
```

Needs changes in PSOL, which are pending review.

### DIFF
--- a/src/ngx_base_fetch.cc
+++ b/src/ngx_base_fetch.cc
@@ -16,6 +16,7 @@
 
 // Author: jefftk@google.com (Jeff Kaufman)
 #include <unistd.h> //for usleep
+#include <inttypes.h> // PRId64
 
 #include "ngx_base_fetch.h"
 #include "ngx_event_connection.h"
@@ -36,7 +37,29 @@ const char kFlush = 'F';
 const char kDone = 'D';
 
 NgxEventConnection* NgxBaseFetch::event_connection = NULL;
-int NgxBaseFetch::active_base_fetches = 0;
+// We'll set this event to uncancelable to prevent nginx from exiting before we
+// are up for it. See ngx_worker_process_cycle().
+ngx_event_t* NgxBaseFetch::shutdown_event = NULL;
+int64 NgxBaseFetch::active_base_fetches = 0;
+int64 NgxBaseFetch::request_ctx_count = 0;
+
+
+// Periodically checks whether the conditions are met for ::Terminate() to be
+// called:
+// - A forced/quick shutdown, e.g. SIGTERM was received.
+// - A graceful shutdown, e.g. SIGQUIT was recevied, after all outstanding
+//   work as been finished.
+// When one of those conditions is met, ev->cancelable will have been set to
+// false by us, and nginx will end up calling ::Terminate() next after calling
+// Done(false) on any outstanding proxy fetches.
+// Terminate() will clear out any pending events to make sure we release all
+// associated NgxBaseFetch instances.
+static void ps_base_fetch_shutdown_event_handler(ngx_event_t *ev) {
+  NgxBaseFetch::CheckShutdownEvent();
+  if (!ev->cancelable) {
+    ngx_add_timer(ev, 1000);
+  }
+}
 
 NgxBaseFetch::NgxBaseFetch(ngx_http_request_t* r,
                            NgxServerContext* server_context,
@@ -65,6 +88,17 @@ NgxBaseFetch::~NgxBaseFetch() {
 bool NgxBaseFetch::Initialize(ngx_cycle_t* cycle) {
   CHECK(event_connection == NULL) << "event connection already set";
   event_connection = new NgxEventConnection(ReadCallback);
+
+  shutdown_event = reinterpret_cast<ngx_event_t*>(
+      ngx_pcalloc(cycle->pool, sizeof(ngx_event_t)));
+  shutdown_event->handler = ps_base_fetch_shutdown_event_handler;
+
+  shutdown_event->data = cycle;
+  shutdown_event->log = cycle->log;
+  // Prevents nginx from exiting until we are up for it.
+  shutdown_event->cancelable = 0;
+  ngx_add_timer(shutdown_event, 1000);
+
   return event_connection->Init(cycle);
 }
 
@@ -72,26 +106,30 @@ void NgxBaseFetch::Terminate() {
   if (event_connection != NULL) {
     GoogleMessageHandler handler;
     PosixTimer timer;
-    int64 timeout_us = Timer::kSecondUs * 30;
+    // A second should be more then enough.
+    int64 timeout_us = Timer::kSecondUs;
     int64 end_us = timer.NowUs() + timeout_us;
     static unsigned int sleep_microseconds = 100;
 
     handler.Message(
-        kInfo,"NgxBaseFetch::Terminate rounding up %d active base fetches.",
+        kInfo,"NgxBaseFetch::Terminate start: %" PRId64 " base fetches.",
         NgxBaseFetch::active_base_fetches);
 
-    // Try to continue processing and get the active base fetch count to 0
-    // untill the timeout expires.
-    // TODO(oschaaf): This needs more work.
+    // Drain any events after a quick or graceful shutdown to clear out
+    // associated NgxBaseFetch instances.
     while (NgxBaseFetch::active_base_fetches > 0 && end_us > timer.NowUs()) {
       event_connection->Drain();
       usleep(sleep_microseconds);
     }
 
-    if (NgxBaseFetch::active_base_fetches != 0) {
+    DCHECK_EQ(NgxBaseFetch::active_base_fetches, 0);
+    DCHECK_EQ(NgxBaseFetch::request_ctx_count, 0);
+
+    if (NgxBaseFetch::active_base_fetches != 0 || NgxBaseFetch::request_ctx_count) {
       handler.Message(
-          kWarning,"NgxBaseFetch::Terminate timed out with %d active base fetches.",
-          NgxBaseFetch::active_base_fetches);
+          kWarning,"NgxBaseFetch::Terminate exit: %" PRId64 " base fetches, "
+          "%" PRId64 " request contexts", NgxBaseFetch::active_base_fetches,
+          NgxBaseFetch::request_ctx_count);
     }
 
     // Close down the named pipe.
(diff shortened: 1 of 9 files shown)


oschaaf:
Tests that ran today brought a problem to light. I'll look into that, and make a comment here when I've pushed the fix.

morlovich:
This looks rather dangerous to me --- the code in ProxyFetch doesn't expect Done to be called externally like this, and it looks like a way towards crashes accessing freed memory, as fetch code may end up still writing towards the fetch object.

oschaaf:
@morlovich thanks for having a look. We call `proxy_fetch->Done(false)` from ngx_pagespeed too [1] in case requests time out, are aborted, or error otherwise, so I thought this would be OK. Should we have a look at that? 

[1] https://github.com/pagespeed/ngx_pagespeed/blob/master/src/ngx_pagespeed.cc#L1696

morlovich:
Does something ensure that you don't call any further Write() and Done()
things?

oschaaf:
@morlovich Yes, nps won't touch any proxy fetches after calling Done(true|false) on them. 
That applies to the proposed CancelOutstanding() as well (the goal here is to be able to shutdown real quick)

morlovich:
Might you not still have an html parse ongoing, though?

oschaaf:
@morlovich Yes, it is possible there are html parses going on indeed. Does calling proxy_fetch->Done(false) not stop/cancel the associated html parser? The code that uses this call is at:
See https://github.com/pagespeed/ngx_pagespeed/pull/936/files#diff-8fb776418fb480e5f084e8500b8f0b0cR3022 (`ps_exit_child_process()`)

This works if we are lucky, but no guarantees. With some synthetic ongoing load this still crashes.
Cancelling out the proxy fetches seemed like a good first step, as we're not going to wait for upstreams at this point and have no further need for them. Any further suggestions? Should there perhaps something be added to the rewrite driver factory too to support a quick shutdown?

morlovich:
Hmm... You're right, it should behave correctly as long as there are no
other Write or Done calls in a sense that the parser's worldview is
consistent.
I was thinking about this completely wrong: you're literally changing the
input events, which is perfectly fine.
(What might matter is that parsing may still go on for a while....)

Sorry for misleading you.

oschaaf:
@morlovich I'm not sure, but would it be a reasonable idea to add a separate call to the driver factory to stop the schedulers as a first step in shutting down quickly?

oschaaf:
@morlovich Running without valgrind, quick shutdowns seems stable after some tweaking. However, I sometimes see sequences like this leading up to a `DCHECK` when running under valgrind:
```
     check [ -z 2015/03/23 00:44:11 [warn] 17691#0: [ngx_pagespeed 1.10.0.0-4597] RateController: drop deferred fetch of http://127.0.0.1:8050/mod_pagespeed_example/images/pagespeed_logo.png on shutdown
2015/03/23 00:44:11 [warn] 17691#0: [ngx_pagespeed 1.10.0.0-4597] RateController: drop fetch of http://127.0.0.1:8050/mod_pagespeed_example/styles/index_style.css on shutdown
2015/03/23 00:44:11 [warn] 17691#0: [ngx_pagespeed 1.10.0.0-4597] RateController: drop fetch of http://127.0.0.1:8050/mod_pagespeed_example/images/pagespeed_logo.png on shutdown
2015/03/23 00:44:11 [warn] 17691#0: [ngx_pagespeed 1.10.0.0-4597] RateController: drop fetch of http://127.0.0.1:8050/mod_pagespeed_example/styles/index_style.css on shutdown
2015/03/23 00:44:11 [warn] 17691#0: [ngx_pagespeed 1.10.0.0-4597] [0323/004412:WARNING:queued_worker_pool.cc(379)] Adding function to sequence 0xf197610 after shutdown
2015/03/23 00:44:11 [warn] 17691#0: [ngx_pagespeed 1.10.0.0-4597] RateController: drop fetch of http://127.0.0.1:8050/mod_pagespeed_example/styles/index_style.css on shutdown
2015/03/23 00:44:11 [alert] 17691#0: [ngx_pagespeed 1.10.0.0-4597] [0323/004412:FATAL:rewrite_driver.cc(2438)] Check failed: !release_driver_. 
``` 

Is this `DCHECK` happening because of something that I overlooked?

With this patch, the quick shutdown looks like this:
1. The nginx side calls `proxy_fetch->Done(false)` on all outstanding proxy fetches
2. The nginx side drains any queued events, and releases all outstanding request context.
3. The nginx side calls `rewrite_driver_factory->ShutDown()`, which should now be the only one holding references to pending drivers/parsers (right?)

@jeffkaufman 
To be able to test this under valgrind, I had to patch nginx because of the box it will force kill child processes one second after forwarding `SIGTERM`, which is too short for us to wrap up under high load. How should this be handled? Would it be OK to patch nginx for testing under valgrind?

jeffkaufman:
Patching nginx to let us finish cleaning up under high load with valgrind sounds like the right way to handle it for me.  Ideally the patch would make shutdown timeout a configuration option, so we could turn it off without rebuiding nginx, but if that's too annoying I can make our scripts just patch the nginx code before running valgrind and handle the rebuild.  If you do make it a configuration option I wonder whether they would accept that upstream?

oschaaf:
@jeffkaufman patch suggested: https://www.mail-archive.com/nginx-devel@nginx.org/msg03473.html

oschaaf:
@morlovich never mind the question about `DCHECK(!release_driver)`.
`ProxyFetch::CancelOutstanding` should mind `done_outstanding_` to make sure it doesn't make a repeated call to Done. Doing so stabilizes the test runs for this.

jeffkaufman:
Is this ready for final review or are you still working on it?

ngx_pagespeed developers would need to patch in your nginx change, right?  When this does go in, can you add that to https://github.com/pagespeed/ngx_pagespeed/wiki/Testing ?

oschaaf:
@jeffkaufman This is ready for review, and I'll make the wiki change once this goes in!

src/ngx_base_fetch.cc
```diff
@@ -193,7 +234,8 @@ ngx_int_t NgxBaseFetch::CopyBufferToNginx(ngx_chain_t** link_ptr) {
   }
 
   int rc = string_piece_to_buffer_chain(
-      request_->pool, buffer_, link_ptr, done_called_ /* send_last_buf */);
+      request_->pool, buffer_, link_ptr, done_called_ /* send_last_buf */,
+      true);
```
jeffkaufman: nit: comment what this "true" means

src/ngx_pagespeed.cc
```diff
@@ -124,7 +126,7 @@ char* string_piece_to_pool_string(ngx_pool_t* pool, StringPiece sp) {
 // (potentially) longer string to nginx and want it to take ownership.
 ngx_int_t string_piece_to_buffer_chain(
     ngx_pool_t* pool, StringPiece sp, ngx_chain_t** link_ptr,
-    bool send_last_buf) {
+    bool send_last_buf, bool flush) {
```
jeffkaufman: nit: for consistency, could we call this `bool send_flush`?

src/ngx_pagespeed.cc
```diff
@@ -1988,9 +1981,22 @@ void ps_send_to_pagespeed(ngx_http_request_t* r,
...
   for (cur = in; cur != NULL; cur = cur->next) {
     last_buf = cur->buf->last_buf;
 
+    if (cur->buf->flush && ctx->follow_flushes) {
+      // Calling ctx->proxy_fetch->Flush(cfg_s->handler) will be a no-op here,
+      // unless we have follow_flushes or flush_html enabled. Note that PSOL
+      // might aggregate multiple flushes into 1, and actually flush a little bit
+      // later due to html parser state and earlier scheduled operations.
+      // Also, unless we also set the flush flag on the nginx buffers we won't
+      // actually flush.
```
jeffkaufman: > unless we also set the flush flag on the nginx buffers we won't actually flush

You added the option to do this when we build the buffer chain, so do we actually flush?

src/ngx_pagespeed.cc
```diff
@@ -1988,9 +1981,22 @@ void ps_send_to_pagespeed(ngx_http_request_t* r,
                           ngx_chain_t* in) {
   ngx_chain_t* cur;
   int last_buf = 0;
+
   for (cur = in; cur != NULL; cur = cur->next) {
     last_buf = cur->buf->last_buf;
 
+    if (cur->buf->flush && ctx->follow_flushes) {
```
jeffkaufman: Does it ever make sense to flush on intermediate buffers in the chain, as opposed to the last buffer?

jeffkaufman:
LGTM

jeffkaufman:
@jmarantz @morlovich do either of you want to look at this more?

src/ngx_pagespeed.cc
oschaaf: When `string_piece_to_buffer_chain` is called by  `NgxBaseFetch` it will always have the `flush` argument set to `true`.
By default, we will never flush at all except when we are done transforming (equalling current behaviour).
We could consider switching the new `follow_flushes` argument to true for nps, which would make us flush when the upstream indicates it want us to. We could recommend enabling `flush_html` for reverse proxy setups.

src/ngx_pagespeed.cc
oschaaf: Perhaps when you need very precise control in chunkiness (for example, streaming / bandwidth throttling)?
This tries to reflect what the upstream wants us to do the best we can.  If the upstream is too chatty, disabling follow_flushes would make sense (this would argue for defaulting that to disabled). In that case, `flush_html` would be a good to enable?

src/ngx_base_fetch.cc
oschaaf: Done (`true` indicates `buf->flush` will be set)

oschaaf:
@jeffkaufman processed comments & force pushed

oschaaf:
Should I make a pull with the PSOL side needed for this to https://github.com/pagespeed/mod_pagespeed ?

jeffkaufman:
> Should I make a pull with the PSOL side needed for this to https://github.com/pagespeed/mod_pagespeed

@oschaaf Yes!  I'm glad to say we can accept pull requests to all of PageSpeed now!

jeffkaufman:
LGTM

oschaaf:
@jeffkaufman  done: https://github.com/pagespeed/mod_pagespeed/pull/1066

PR #943, opened 2015-03-30
move 'type' variable to inside an #if statement

crowell:
removes an unused variable when compile without NGX_DEBUG, allows for
building with -Werror.

### DIFF
--- a/src/ngx_base_fetch.cc
+++ b/src/ngx_base_fetch.cc
@@ -120,14 +120,16 @@ void NgxBaseFetch::ReadCallback(const ps_event_data& data) {
   NgxBaseFetch* base_fetch = reinterpret_cast<NgxBaseFetch*>(data.sender);
   ngx_http_request_t* r = base_fetch->request();
   bool detached = base_fetch->detached();
+#if (NGX_DEBUG)  // `type` is unused if NGX_DEBUG isn't set, needed for -Werror.
   const char* type = BaseFetchTypeToCStr(base_fetch->base_fetch_type_);
+#endif
   int refcount = base_fetch->DecrementRefCount();
 
-  #if (NGX_DEBUG)
+#if (NGX_DEBUG)
   ngx_log_error(NGX_LOG_DEBUG, ngx_cycle->log, 0,
      "pagespeed [%p] event: %c. bf:%p (%s) - refcnt:%d - det: %c", r,
      data.type, base_fetch, type, refcount, detached ? 'Y': 'N');
-  #endif
+#endif
 
   // If we ended up destructing the base fetch, or the request context is
   // detached, skip this event.


src/ngx_base_fetch.cc
```diff
@@ -120,10 +120,10 @@ void NgxBaseFetch::ReadCallback(const ps_event_data& data) {
   NgxBaseFetch* base_fetch = reinterpret_cast<NgxBaseFetch*>(data.sender);
   ngx_http_request_t* r = base_fetch->request();
   bool detached = base_fetch->detached();
-  const char* type = BaseFetchTypeToCStr(base_fetch->base_fetch_type_);
   int refcount = base_fetch->DecrementRefCount();
 
   #if (NGX_DEBUG)
+  const char* type = BaseFetchTypeToCStr(base_fetch->base_fetch_type_);
```
jeffkaufman: You can't decrement the refcount and then call base_fetch->base_fetch_type_

crowell:
moved into separate `#else` block.

src/ngx_base_fetch.cc
```diff
@@ -120,13 +120,15 @@ void NgxBaseFetch::ReadCallback(const ps_event_data& data) {
...
-  int refcount = base_fetch->DecrementRefCount();
 
   #if (NGX_DEBUG)
+  const char* type = BaseFetchTypeToCStr(base_fetch->base_fetch_type_);
+  int refcount = base_fetch->DecrementRefCount();
   ngx_log_error(NGX_LOG_DEBUG, ngx_cycle->log, 0,
      "pagespeed [%p] event: %c. bf:%p (%s) - refcnt:%d - det: %c", r,
      data.type, base_fetch, type, refcount, detached ? 'Y': 'N');
+  #else
+  int refcount = base_fetch->DecrementRefCount();
```
jeffkaufman: Can't we just move `DecrementRefCount()` after the NGX_DEBUG section and avoid the  duplication?

crowell: refcount is used in the log message.
I could just wrap the type declaration in an `#if (NGX_DEBUG)` which might make more sense/no duplication.

crowell: alternative way:
```cpp
void NgxBaseFetch::ReadCallback(const ps_event_data& data) {                    
  NgxBaseFetch* base_fetch = reinterpret_cast<NgxBaseFetch*>(data.sender);      
  ngx_http_request_t* r = base_fetch->request();                                
  bool detached = base_fetch->detached();                                       
#if (NGX_DEBUG)  // `type` is unused if NGX_DEBUG isn't set, needed for -Werror.
  const char* type = BaseFetchTypeToCStr(base_fetch->base_fetch_type_);         
#endif                                                                          
  int refcount = base_fetch->DecrementRefCount();                               
                                                                                
#if (NGX_DEBUG)                                                                 
  ngx_log_error(NGX_LOG_DEBUG, ngx_cycle->log, 0,                               
     "pagespeed [%p] event: %c. bf:%p (%s) - refcnt:%d - det: %c", r,           
     data.type, base_fetch, type, refcount, detached ? 'Y': 'N');               
#else                                                                           
```

jeffkaufman: Gah, awkwardness.

Yes, wrapping the `type` sounds ok.

crowell:
@jeffkaufman squashed and pushed.

I updated the `#if` `#endif` spacing to match the other usage in this file, and to match the clang-format style.

jeffkaufman:
LGTM

PR #944, opened 2015-03-30
Adjust to API change to InitiatePropertyCacheLookup

morlovich:
The last argument was not actually used, and hence was removed.

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -344,7 +344,7 @@ ngx_int_t ps_base_fetch_handler(ngx_http_request_t* r) {
 
   if (rc == NGX_ERROR) {
     ps_set_buffered(r, false);
-    ps_release_base_fetch(ctx);    
+    ps_release_base_fetch(ctx);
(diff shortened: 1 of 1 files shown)


oschaaf:
LGTM

src/ngx_pagespeed.cc
```diff
@@ -1921,8 +1920,7 @@ ngx_int_t ps_resource_handler(ngx_http_request_t* r,
             cfg_s->server_context,
             options,
             ctx->base_fetch,
-            false /* requires_blink_cohort (no longer unused) */,
-            &page_callback_added);
+            false /* requires_blink_cohort (no longer unused) */);
```
jeffkaufman: Is "no longer unused" correct?  The double negative is a bit weird, hinting at "unused" or "no longer used" instead of the literal "now used".

jeffkaufman:
LGTM

src/ngx_pagespeed.cc
morlovich: proxy_interface does pass true for the last argument if cache_html is on, so that argument isn't dead, at least.

PR #966, opened 2015-05-14
Handle DoneAndSetHeaders() new second argument that wants to know whe…

jeffkaufman:
…ther the response is complete.

Motivated by the PSOL change to fix https://github.com/pagespeed/mod_pagespeed/issues/1081


oschaaf:
Great, LGTM!

PR #967, opened 2015-05-18
nginx system test changes to repro issue https://github.com/pagespeed…

jmarantz:
…/mod_pagespeed/issues/1077


crowell:
LGTM

@jeffkaufman how do you get googlebot to ignore pull requests from people?

jeffkaufman:
@crowell We can't yet; that feature is still coming.

PR #989, opened 2015-07-14
headers: only set Last-Modified where we support If-Modified-Since

jeffkaufman:


oschaaf:
LGTM

jeffkaufman:
mod_pagespeed eb95331 is in, so submitting the nginx half

Issue #903, opened 2015-02-09
Allow DownstreamCachePurgeLocation to use Nginx script variables 


jaseclamp:
As defined in this stackoverflow question: 

http://stackoverflow.com/questions/28355890/ngx-pagespeed-behind-varnish-how-specify-multiple-or-correct-downstreamcachep/28356567#28356567

Currently ProcessScriptVariables only applies to LoadFromFileXXX but expanding that to also include DownstreamCachePurgeLocation would allow Varnish to pass on to Nginx (via headers) the IP address of which downstream cache routed the traffic and consequently PageSpeed will be able to purge the correct Cache. 

Thanks

franklinwise:
+1

jeffkaufman:
Could you try the `allow-downstream-caching-scripting` branch and check that it does what you want?

Installation instructions:

```
cd
NPS_VERSION=1.9.32.4
git clone git@github.com:pagespeed/ngx_pagespeed.git
cd ngx_pagespeed
git checkout allow-downstream-caching-scripting
wget https://dl.google.com/dl/page-speed/psol/${NPS_VERSION}.tar.gz
tar -xzvf ${NPS_VERSION}.tar.gz  # extracts to psol/

cd
NGINX_VERSION=1.8.0
wget http://nginx.org/download/nginx-${NGINX_VERSION}.tar.gz
tar -xvzf nginx-${NGINX_VERSION}.tar.gz
cd nginx-${NGINX_VERSION}/
./configure --add-module=$HOME/ngx_pagespeed
make
sudo make install
```

This is almost the same as https://developers.google.com/speed/pagespeed/module/build_ngx_pagespeed_from_source , but tweaked to use a git checkout instead of a downloaded release.

PR #994, opened 2015-07-22
config: whitelist a few more directives for script variables

jeffkaufman:
When configuring downstream caching if you have an origin with multiple caches in front of it you might want to choose how you handle requests in response to which cache you're getting them from.  White-list `DownstreamCachePurgeLocationPrefix`, `DownstreamCachePurgeMethod`, and `DownstreamCacheRewrittenPercentageThreshold` for scripting so this is possible.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/903


oschaaf:
LGTM

PR #1002, opened 2015-07-31
ignore errors about slow file operations

jmarantz:
Slow file operations can result now in error logs, and those can occur when running valgrind tests.  We need to ignore those in a grep for errors in the error log.


crowell:
LGTM

Issue #957, opened 2015-04-27
defer_javascript / js_defer not loaded (0 bytes) when using pagespeed in combination with HTTPS


marcoroest87:
When using ngx_pagespeed in combination HTTPS the js_defer.js isn't loaded on the first request (SSL negotiation?)

I'm running nginx 1.7.12 + ngx_pagespeed (master branch, but I tried 1.9.32.3 with the same result)  on debian wheezy. 

My setup looks like this 
1. Nginx (443) - pagespeed and SSL offloading
2. Varnish (8081)
3 Nginx (8080)

This is easily reproducable by using Firefox after hard reloading the page. This can be tested here: https://demo.hypershop.nl/activity-trackers.html (on all other pages the defer_javascript filter is disabled). The issue also occurs using Chrome, but less often is seems.
Since we load the main css with javascript (top css gets included into the html) this results in a page where only the top is shown and no javascript is loaded :(

There are no errors in nginx's error log. (I've tried buidling Nginx with --with-debug option, but this resulted in nginx not being able to server any pages at all).

This is what the access log shows:
    94.209.44.170 - - [27/Apr/2015:16:59:36 +0200] "GET /pagespeed_static/1.JiBnMqyl6S.gif HTTP/1.1" 200 0 "https://demo.hypershop.nl/activity-trackers.html" "Mozilla/5.0 (Windows NT 6.1; WOW64; rv:37.0) Gecko/20100101 Firefox/37.0"

Screenshot (Firefox):
https://demo.hypershop.nl/firefox.jpg

This i've tried: 
* using LoadFromFile instead of using FetchHttps
* Not using SPDY (using SSL instead)
* tried adding the filters in different orders
* Using a HTTP proxy like Fiddler. When using Fiddlers ssl certificate in Firefox I cannot reproduce this issue

Config files: 
* Nginx proxy (80): https://gist.github.com/marcoroest87/73bb1f1c48dfef5d905c
* pagespeed.conf  https://gist.github.com/marcoroest87/6673175923a7e45fecd3
* ssl.conf https://gist.github.com/marcoroest87/584fd8faef4ca31087a1
* Nginx backend (8080): https://gist.github.com/marcoroest87/6abbc6661828ce3c4ccf

This is how I've build nginx:
	./configure \
		--prefix=/usr/share/nginx \
		--sbin-path=/usr/sbin/nginx \
		--conf-path=/etc/nginx/nginx.conf \
[...]

PR #1014, opened 2015-10-06
simple_handler: pass through error codes

oschaaf:
For https://github.com/pagespeed/ngx_pagespeed/issues/957
I can't reproduce the problem on my machine yet, but I think it is
worth trying to see if this change fixes the issue.


oschaaf:
whoops, accidental pull request, closing. Sorry!

PR #1018, opened 2015-10-10
Fixes pagespeed/ngx_pagespeed#957

rfnx:


rfnx:
I signed it ayy

oschaaf:
Awesome. This change looks good to me, but I want to manually test a few things before I would be happy with it being merged. I expect to get back on that next monday.
Thanks!

jeffkaufman:
This makes sense to me.  LGTM.

@oschaaf let me know when you're comfortable with it being merged?

oschaaf:
@jeffkaufman LGTM, I'm comfortable with merging. 

Note: one system test fails on me, `Cache purge test`, but I can't relate that to this change.

PR #1026, opened 2015-10-22
Don't call RewriteDriver::SetUserAgent on the UA in the request-headers.

jmarantz:
RewriteDriver::SetRequestHeaders already pulls the UA from the request-headers.


jeffkaufman:
LGTM

PR #1030, opened 2015-10-29
Huibao webp

huibaolin:
Add two folders for testing WebP images under /mod_pagespeed_test/optimize_for_bandwidth/.


jeffkaufman:
@huibaolin It looks like you made this pull request against `master`, but it should have gone against `trunk-tracking`.  That's why it looks like so many files are changed.  Could you delete it, and make a new one against `trunk-tracking`?

PR #1046, opened 2015-11-16
add content_encoding to headers_out

crowell:
this change is required for gzip cache change going into mod_pagespeed.

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -607,6 +607,16 @@ ngx_int_t copy_response_headers_to_ngx(
       CHECK(pagespeed_headers.FindContentLength(&len));
       headers_out->content_length_n = len;
       headers_out->content_length = header;
+    } else if (STR_EQ_LITERAL(name, "Content-Encoding")) {
+      headers_out->content_encoding = header;
+    } else if (STR_EQ_LITERAL(name, "Refresh")) {
+      headers_out->refresh = header;
+    } else if (STR_EQ_LITERAL(name, "Content-Range")) {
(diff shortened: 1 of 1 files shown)


crowell:
@jeffkaufman this change unblocks the gzipped-cache change and allows it to go it.

jeffkaufman:
The shortcuts in `headers_out_t` are:
```
server, date, content_length, content_encoding, location, refresh, last_modified,
content_range, accept_ranges, www_authenticate, expires, and etag
```

With this change we'll be populating:
```
Date, Etag, Expires, Last-Modified, Location, Server,
Content-Length, and Content-Encoding.
```

That leaves:
```
refresh, content_range, accept_ranges, and www_authenticate.
```

Any reason not to add these too, so some future person generating a new header in PSOL doesn't get bitten?

crowell:
additional shortcuts added

jeffkaufman:
LGTM

src/ngx_pagespeed.cc
```diff
@@ -607,6 +607,16 @@ ngx_int_t copy_response_headers_to_ngx(
       CHECK(pagespeed_headers.FindContentLength(&len));
       headers_out->content_length_n = len;
       headers_out->content_length = header;
+    } else if (STR_EQ_LITERAL(name, "Content-Encoding")) {
+      headers_out->content_encoding = header;
```
oschaaf: I wonder if we should look at `r->gzip_vary` and any pre-existing `Vary: Accept-Encoding` headers here. Does PSOL also add a `Vary` header when it emits a `Content-Encoding` header?

PR #1045, opened 2015-11-14
Fix interaction with ngx_brotli module.

PiotrSikora:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1021

PR #1047, opened 2015-11-16
Fix interaction with ngx_brotli module.

jeffkaufman:
Pull @PiotrSikora 's fix from #1045 onto trunk-tracking.

This did mean fixing up some merge conflicts: trunk-tracking (which is newer than master) has a `position_aux` option.


PiotrSikora:
LGTM.

crowell:
LGTM

Issue #942, opened 2015-03-26
Builds failing in GCC 5/new ABI


kyl191:
I'm having issues building ngx_pagespeed on Fedora Rawhide (aka Fedora 23), but not Fedora 19-22. I suspect the issue is steming from [GCC 5's ABI change](http://developerblog.redhat.com/2015/02/05/gcc5-and-the-c11-abi/).

Full output of obj/autoconf.err: http://paste.fedoraproject.org/203083/27353406/, what I suspect is the relevant error below:

    checking for psol
    /tmp/ccWlcgjO.o: In function `base::BasicStringPiece<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::BasicStringPiece(char const*)':
    autotest.cc:(.text._ZN4base16BasicStringPieceINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEC2EPKc[_ZN4base16BasicStringPieceINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEC5EPKc]+0x1f): undefined reference to `base::internal::StringPieceDetail<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::StringPieceDetail(char const*)'
    /tmp/ccWlcgjO.o: In function `net_instaweb::HtmlParse::ParseText(base::BasicStringPiece<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > const&)':
    autotest.cc:(.text._ZN12net_instaweb9HtmlParse9ParseTextERKN4base16BasicStringPieceINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEE[_ZN12net_instaweb9HtmlParse9ParseTextERKN4base16BasicStringPieceINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEEE]+0x29): undefined reference to `base::internal::StringPieceDetail<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::size() const'
[...]

PR #1048, opened 2015-11-24
Pass in -D_GLIBCXX_USE_CXX11_ABI=0, so people using gcc-5.x can use our psol.a

morlovich:
(built with 4.8)

See issue 942


morlovich:
Hmm, might need a long link?
https://github.com/pagespeed/ngx_pagespeed/issues/942

crowell:
lgtm

PR #1050, opened 2015-11-25
Add compiler version checks for what we want to support going forward,

morlovich:
and make us actually build with clang (the log_message_handler.cc change
and config.make change)

### DIFF
--- a/config
+++ b/config
@@ -71,6 +71,20 @@ else
   buildtype=Release
 fi
 
+# If the compiler is gcc, we want to use g++ to link, if at all possible,
+# so that -static-libstdc++ works.
+# Annoyingly, the feature test doesn't even use $LINK for linking, so that
+# needs an explicit -lstdc++
+pagespeed_libs=
+ps_maybe_gpp_base=`basename $CC| sed s/gcc/g++/`
+ps_maybe_gpp="`dirname $CC`/$ps_maybe_gpp_base"
+if [ -n "$NGX_GCC_VER" -a \( -x "$ps_maybe_gpp" \) ]; then
+  LINK=$ps_maybe_gpp
+  NGX_TEST_LD_OPT="$NGX_TEST_LD_OPT -lstdc++"
+else
+  pagespeed_libs="-lstdc++"
+fi
+
 # The compiler needs to know that __sync_add_and_fetch_4 is ok,
 # and this requires an instruction that didn't exist on i586 or i386.
 if [ "$uname_arch" = "i686" ]; then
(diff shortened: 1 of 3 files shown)


morlovich:
@jeffkaufman any chance you could bump this in your queue? It's sort of release-blocking, after all...

config
```diff
@@ -234,4 +251,46 @@ END
...
+ngx_feature_incs=""
+ngx_feature_path=""
+ngx_feature_libs="-lstdc++"
+ngx_feature_test="
+
+#if defined(__clang__) && defined(__GLIBCXX__)
+   // See https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html#abi.versioning 
+   // for a list of various values of __GLIBCXX__. Note that they're not monotonic
+   // with respect to version numbers.
+   #if __GLIBCXX__ == 20120322  ||__GLIBCXX__ == 20120614
```
jeffkaufman: nit: whitespace around the `||`

config
```diff
@@ -234,4 +251,46 @@ END
...
+      #error \"clang is using libstdc++ 4.7.0 or 4.7.1, which can cause binary incompatibility.\"
+   #endif
+#endif
+
+#if !defined(__clang__) && defined(__GNUC__)
+   #if __GNUC__ < 4 || (__GNUC__ == 4 && __GNUC_MINOR__ < 8)
+      #error \"GCC < 4.8 no longer supported. Please use gcc >= 4.8 or clang >= 3.3\"
```
jeffkaufman: We have instructions for how to do this easily, right?  Can this generate a link to them?

config
```diff
@@ -234,4 +251,46 @@ END
...
+   #endif
+#endif
+
+#if defined(__clang__)
+   #if __clang_major__ < 3 || (__clang_major__ == 3 && __clang_minor__ < 3)
+      #error \"Please use gcc >= 4.8 or clang >= 3.3\"
```
jeffkaufman: same here

jeffkaufman:
aside from minor comments, LGTM

config
morlovich: See line 289 below. That's what's ultimately printed to the user. This is what goes to autoconf.err

config
morlovich: Good point. Now I just need to remember where I put this. 
Oh, I guess I could just pull this, can't I?

jeffkaufman: You can just check out the repo with git, check out the morlovich-compiler-req, make your edit, and push back up.

config
jeffkaufman: Sounds good.

morlovich:
Should I merge this, Jeff?

jeffkaufman:
Yes, go ahead and merge.

Issue #819, opened 2014-10-02
pagespeed changes location response header in a 302 redirect


demonccc:
Hi,

I'm using HAproxy as loadbalancer for some Nginx instances and it provides a HTTPS secure connection to the client. 

The Nginx instances are configured to listen on the port 80 (HTTP). The PHP application create a redirect response using an relative path as URL, but when the response is retrieved by the browser it doesn't work correctly because the location header of the response was replaced by an absolute URL with HTTP. 

Is it a bug or there are some parameters to set in order to avoid this issue? I tried to find a solution for two weeks, but I didn't find anything.

Regards.

Issue #711, opened 2014-05-28
PageSpeed redirects to port 8080 with downstream cache configuration


nikolay:
I've reported this along with other issues, but want to put this is a top-level issue now.

I have WordPress configured with Downstream Caching. So Ajax requests (such as post search) incorrectly rewrite the WordPress redirects to have port 8080, which is only used internally behind the caching layer.

This issues is not present in versions before 1.7.

It's a critical issue for us, I've burned a lot of hours trying to work around the issue, but nothing seems to work. Please, help! Seriously!

Issue #1029, opened 2015-10-28
RespectXForwardedProto not working properly


mlimaloureiro:
Hi this issue is very similar to this one https://github.com/pagespeed/ngx_pagespeed/issues/819.

My env consists in application servers (ec2 instances) behind an ELB where the ssl termination occur. So the HTTPS request is made to the load balancer, which then internally requests the EC2 instance over HTTP.

An example of what's happening is:

1. user makes a POST request over HTTPS.
2. the request reaches the application server over HTTP and is handled.
3. the redirect should go to an HTTPS url but it's coming in HTTP.

By reading the docs 

```
 pagespeed RespectXForwardedProto on;
```

should fix this issue but it's not working in my case. If I turn off pagespeed everything works well.

I currently have this in my config with pagespeed 1.9.32.10 version

```
    pagespeed on;
    pagespeed FileCachePath /var/ngx_pagespeed_cache;
    pagespeed EnableFilters lazyload_images;
    pagespeed CriticalImagesBeaconEnabled false;
    pagespeed RespectXForwardedProto on;
    pagespeed FileCacheSizeKb            153600;  # 150 MB
    pagespeed FileCacheCleanIntervalMs  3600000;  # 1 hour
    pagespeed FileCacheInodeLimit        500000;

    location ~ "\.pagespeed\.([a-z]\.)?[a-z]{2}\.[^.]{10}\.[^.]+" {
        add_header "" "";
    }
    location ~ "^/pagespeed_static/" { }
    location ~ "^/ngx_pagespeed_beacon$" { }
    location /ngx_pagespeed_statistics { allow 127.0.0.1; deny all; }
    location /ngx_pagespeed_message { allow 127.0.0.1; deny all; }
    location /pagespeed_console { allow 127.0.0.1; deny all; }
```

any help on this would be appreciated.

oschaaf:
@mlimaloureiro Could you try patching in https://github.com/pagespeed/ngx_pagespeed/commit/26a0c37b973a532b18f01c54f79598bd66bb546b to see if that fixes the problem?

mlimaloureiro:
thanks @oschaaf going to check

mlimaloureiro:
@oschaaf yep, works perfectly when patching https://github.com/pagespeed/ngx_pagespeed/commit/26a0c37b973a532b18f01c54f79598bd66bb546b any reason for this not being merged ?

oschaaf:
[...]

PR #1055, opened 2015-11-30
location-header: Be careful with headers_out->location

oschaaf:
Only set headers_out->location when the upstream originally did
as well. If the Location: header value involved starts with "/"
nginx will absolutify it, ignoring any X-Forwarded-Proto header
in the process.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/819
(Confirmed: https://github.com/pagespeed/ngx_pagespeed/issues/1029)
Hopefully fixes https://github.com/pagespeed/ngx_pagespeed/issues/711

### DIFF
--- a/src/ngx_pagespeed.h
+++ b/src/ngx_pagespeed.h
@@ -103,6 +103,11 @@ typedef struct {
   // We need to remember the URL here as well since we may modify what NGX
   // gets by stripping our special query params and honoring X-Forwarded-Proto.
   GoogleString url_string;
+
+  // We need to remember if the upstream had headers_out->location set, because
+  // we should mirror that when we write it back. nginx may absolutify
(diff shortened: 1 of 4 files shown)


src/ngx_pagespeed.h
```diff
@@ -103,6 +103,7 @@ typedef struct {
   // We need to remember the URL here as well since we may modify what NGX
   // gets by stripping our special query params and honoring X-Forwarded-Proto.
   GoogleString url_string;
+  bool location_field_set;
```
jeffkaufman: Needs comment.

jeffkaufman:
Aside from missing comment, LGTM

src/ngx_pagespeed.h
oschaaf: Missing comment added

Issue #895, opened 2015-02-01
Log No threading detected is back in 1.9.32.3-4448


PieterGit:
According to issue #808 the logging of errors such as:

2015/01/31 22:37:25 [info] 4939#0: [ngx_pagespeed 1.9.32.3-4448] No threading detected. Own threads: 1 Rewrite, 1 Expensive Rewrite.
2015/02/01 08:46:34 [info] 3179#0: [ngx_pagespeed 1.9.32.3-4448] No threading detected. Own threads: 1 Rewrite, 1 Expensive Rewrite.
2015/02/01 08:49:04 [info] 3218#0: [ngx_pagespeed 1.9.32.3-4448] No threading detected. Own threads: 1 Rewrite, 1 Expensive Rewrite.
2015/02/01 08:53:10 [info] 3274#0: [ngx_pagespeed 1.9.32.3-4448] No threading detected. Own threads: 1 Rewrite, 1 Expensive Rewrite.

Should be fixed with 1.9.32.2. Unfortuately it's seems back with 1.9.32.3-4448.

Please degrade this message to debug level, because it's confusing and its logged in 
/var/log/nginx/error.log and confused me.

jeffkaufman:
Issue #808 was for this message being printed even if people had set their log level to exclude info messages.  That was fixed with e2e21474cea3 and was included in `1.9.32.3` as well as `1.9.32.2` and `master`.

You're saying this should be a `debug` message instead of `info`, and I agree.  I'll downgrade it.

jeffkaufman:
Looking at the code, `kInfo` is actually our noisiest debug level; we don't have a `kDebug`.  Is there a reason you need to run your server with Info logging on?

PieterGit:
I do not run my server with 'info' logging on.

I have the following log settings in my configuration : ("grep -i log *.conf")
nginx.conf:     # Logging Settings
nginx.conf:     access_log /var/log/nginx/access.log;
nginx.conf:     error_log /var/log/nginx/error.log;

According to http://nginx.org/en/docs/ngx_core_module.html#error_log
the default level of error_log is error, which does not explain why I see these messages in the error_log. I will now increase the error_log to "crit" to see if the messages disappear.

My version information:
$ /sbin/nginx -V
nginx version: nginx/1.6.2
built by gcc 4.8.2 (Ubuntu 4.8.2-19ubuntu1)
TLS SNI support enabled
[...]

PR #1071, opened 2015-12-11
log: initialize logging earlier

oschaaf:
It turns out to be possible to initialize logging earlier by
grabbing the log from a global ngx_cycle structure.

This makes us start logging earlier, yet loses the
"No threading detected ..." messages both from stderr and
in error.log when nginx initially starts.

With this change, these messages will now be logged as we start
logging earlier:

"
flush
.
"

These originate from SystemCachePath::CacheKey which appends
newlines to the key, and the resulting cache key ends up being
logged. We might want to change that (on the mps side), because the resulting
lines in error.log look weird and might raise questions.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/895


jeffkaufman:
LGTM

Looks good.  I'll fix the cache path key issue having newlines on the mps side.

PR #1072, opened 2015-12-12
Deleted the duplicate bracket

nikolay:


oschaaf:
LGTM

@jeffkaufman To float an idea: Would it be possible to pull the capabilities list out of PSOL based on the incoming user-agent? If so, I wonder if we could append a request-header based on that and use that header in configuration when building the cache key - instead of duplicating the capability metadata/computations.

jeffkaufman:
@nikolay Good catch!

Could you close this pull request and open a new one against the trunk-tracking branch?  As it is this wants to pull in a lot of commits that aren't yours that we shouldn't move over.

PR #1076, opened 2015-12-15
Store whether h2 is in use in RequestContext.

morlovich:


oschaaf:
A system test would be nice, but I wouldn't know how to implement it today without putting in an unreasonable effort. LGTM.

morlovich:
Well, right now it is especially hard to test since nothing looks at this bit at all. I hand-tested for adding a debug statement.

Eventually we will probably have some tests that expect nghttp to be in $PATH or something, and disable if it's not there --- sadly I don't think command-line h2 clients are yet widely available. Hmm, even our google-side test environment will only build it when also building Apache 2.4; I should probably modify that to build the nghttp command-line client even if Apache doesn't have h2 support so it's available for testing of nginx.

PR #1085, opened 2015-12-24
Fix crasher on 404 .pagespeed. resources w/a custom location

### LINKED ISSUE #1081: core dumps and slow reads with 1.10.33.1-7580 (nginx freezes after some hours)
I recently upgraded to 1.10.33.1, but I had to switch back to 1.9.32.10-7423 (last version i was using) due to following errors, that caused my nginx to freeze after few hours of high traffic (nginx daemon didn't died, but was unable to serve any request)
```
2015/12/21 13:15:02 [error] 29209#0: [ngx_pagespeed 1.10.33.1-7580] Slow ReadFile operation on file /home/pagespeed_cache..omissis..t/it/_.pagespeed.jo.RAuz1oeDfS.js,: 68.457ms; configure SlowFileLatencyUs to change threshold

2015/12/21 13:15:02 [error] 29209#0: [ngx_pagespeed 1.10.33.1-7580] Slow write operation on file /home/pagespeed_cache/....omissis...

2015/12/21 13:16:52 [alert] 11520#0: worker process 29526 exited on signal 11 (core dumped)
2015/12/21 13:17:22 [alert] 11520#0: worker process 29491 exited on signal 11 (core dumped)
2015/12/21 13:20:48 [alert] 11520#0: worker process 29558 exited on signal 11 (core dumped)
2015/12/21 13:22:36 [alert] 11520#0: worker process 29209 exited on signal 11 (core dumped)
2015/12/21 13:23:04 [alert] 11520#0: worker process 29504 exited on signal 11 (core dumped)
2015/12/21 13:23:40 [alert] 11520#0: worker process 29501 exited on signal 11 (core dumped)
2015/12/21 13:33:48 [alert] 11520#0: worker process 29673 exited on signal 11 (core dumped)
2015/12/21 13:34:18 [alert] 11520#0: worker process 29812 exited on signal 11 (core dumped)
2015/12/21 13:34:18 [alert] 11520#0: worker process 29931 exited on signal 11 (core dumped)
2015/12/21 13:34:27 [alert] 11520#0: worker process 29917 exited on signal 11 (core dumped)
2015/12/21 13:34:29 [alert] 11520#0: worker process 30298 exited on signal 11 (core dumped)
2015/12/21 13:34:38 [alert] 11520#0: worker process 30327 exited on signal 11 (core dumped)
```

I'm using nginx 1.8.0 but i tried also to go with 1.9.9, with no luck, now I think something new in pagespeed is the issue for the signal 11 issue as I'm not getting it anymore after switching back to 1.9.32.10

oschaaf:
- Fix nginx-side flow so we handle .pagespeed. resources ok
  when they will land on a customized 404 internal location.
- Additionally, check for a wiped request context and make sure
  we do not dereference a null pointer, which is what hurt in
  the flow we entered above as the IPRO lookup still was
  generating events while the nginx side request context was
  gone.
- Also, as a preliminary measure, do not check fail when we
  receive a stale event originating from a NgxBaseFetch that
  is no longer associated with our request context.
  Do log a warning so we'll hear about this happening either
  through system test failures or a bug report.

Fixes #1081

### DIFF
--- a/src/ngx_base_fetch.cc
+++ b/src/ngx_base_fetch.cc
@@ -118,6 +118,8 @@ const char* BaseFetchTypeToCStr(NgxBaseFetchType type) {
   return "can't get here";
 }
 
+// TODO(oschaaf): replace the ngx_log_error with VLOGS or pass in a
+// MessageHandler and use that.
 void NgxBaseFetch::ReadCallback(const ps_event_data& data) {
   NgxBaseFetch* base_fetch = reinterpret_cast<NgxBaseFetch*>(data.sender);
   ngx_http_request_t* r = base_fetch->request();
@@ -138,9 +140,30 @@ void NgxBaseFetch::ReadCallback(const ps_event_data& data) {
   if (refcount == 0 || detached) {
     return;
   }
+
   ps_request_ctx_t* ctx = ps_get_request_context(r);
 
-  CHECK(data.sender == ctx->base_fetch);
+  // If our request context was zeroed, skip this event.
+  // See https://github.com/pagespeed/ngx_pagespeed/issues/1081
+  if (ctx == NULL) {
+    // Should not happen normally, when it does this message will cause our
+    // system tests to fail.
+    ngx_log_error(NGX_LOG_WARN, ngx_cycle->log, 0,
+        "pagespeed [%p] skipping event: request context gone", r);
+    return;
+  }
+
+  // Normally we expect the sender to equal the active NgxBaseFetch instance.
+  DCHECK(data.sender == ctx->base_fetch);
+
+  // If someone changed our request context or NgxBaseFetch, skip processing.
+  if (data.sender != ctx->base_fetch) {
+      ngx_log_error(NGX_LOG_WARN, ngx_cycle->log, 0,
+          "pagespeed [%p] skipping event: event originating from disassociated"
+          " NgxBaseFetch instance.", r);
+      return;
+  }
+
   CHECK(r->count > 0) << "r->count: " << r->count;
 
   int rc;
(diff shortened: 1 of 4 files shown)


creativeprogramming:
I confirm that this solves the issue, please merge

oschaaf:
If anyone is up for reviewing, this pull is ready for it: I pushed my last changes to this, adding a test and logging warnings.

src/ngx_base_fetch.cc
```diff
@@ -138,9 +140,30 @@ void NgxBaseFetch::ReadCallback(const ps_event_data& data) {
...
 
-  CHECK(data.sender == ctx->base_fetch);
+  // If our request context was zeroed, skip this event.
+  // See https://github.com/pagespeed/ngx_pagespeed/issues/1081
+  if (ctx == NULL) {
+    // Should not happen normally, when it does this message will cause our
+    // system tests to fail.
+    ngx_log_error(NGX_LOG_WARN, ngx_cycle->log, 0,
+        "pagespeed [%p] skipping event: request context gone", r);
+    return;
```
jmarantz: I'm not sure of the semantics of this function and how it related to event handling in nginx.  But I'm concerned that just logging an error and returning will leave the system in a weird state.  Don't we need to call a callback or set a 'did not handle' flag somehow?

oschaaf: This function processes events from pagespeed as we receive them through the named pipe, and subsequently drives nginx's request processing.  When the condition handled here arises, we have received an event originating from an `NgxBaseFetch` instance that is no longer associated to the request it originated from (the whole request context has been zeroed). 

Amongst other cases, nginx zeroes the active request context when it performs an internal redirect to serve a custom error page.

There was a flow/programming error earlier on the nginx side specific to .pagespeed. 404 responses which resolved a custom error page through such an internal redirect. I fixed that, and now I'm not able to reproduce this condition at all in any flow I can think of. 

So the changes for this function are about being more defensive against mistakes and making sure we fail the system tests on a bad request execution flow. For running production, my thought is that CHECK failing would take out the server, while not processing the event *might* take down a request. Processing an event that should not have been processed may make us output things we don't want, whereas not processing it when it should be may stall a request, which seems the better option.
We do log a warning about it, because when this happens we should take a look at it.

jeffkaufman:
LGTM

deluxor:
Saw some crashes from Master, now its perfect:
 GitHub Logo
Format: Alt Text

PR #1088, opened 2015-12-31
reapply:  location-header: tweak location header handling

### LINKED ISSUE #725: Duplicate Location headers
This issue was introduced in 1.8.31.2
With nginx running as a reverse proxy, if the upstream server responds with a 301 response, an extra Location header is added to the response.
This causes the browser (Chrome) to display an 'Duplicate Headers received from Server' error page.

crowell:
- Fix potentially sending the location header into PSOL twice.
- Be more thorough when unsetting the location header

Attempts to fix #725

from oschaaf


oschaaf:
LGTM

crowell:
this also needs to be applied to trunk-tracking

PR #1092, opened 2016-01-05
remove 'IPRO flow uses cache as expected' from flaky-list, ...

jmarantz:
but strip out 'Fetch timed out' from log-warning check


oschaaf:
LGTM

Issue #1096, opened 2016-01-20
IPRO check-fails on some connection errors


grayfolk:
When i turn pagespeed on then have aborted connection and this error.log message:
```
2016/01/19 20:11:55 [alert] 26489#0: [ngx_pagespeed 1.10.33.2-7600] [0119/201155:FATAL:ngx_base_fetch.cc(144)] Check failed: r->count > 0. r->count: 0
nginx: worker process() [0x813edcb]
nginx: worker process() [0x80e16f0]
nginx: worker process() [0x813b9ac]
nginx: worker process() [0x80e2638]
nginx: worker process() [0x80e3102]
nginx: worker process() [0x80e3173]
nginx: worker process() [0x806b8bc]
nginx: worker process() [0x806b48a]
nginx: worker process() [0x80711f2]
nginx: worker process() [0x806fb0f]
nginx: worker process() [0x8072282]
nginx: worker process() [0x8052bd2]
/lib/libc.so.6(__libc_start_main+0xe6) [0x795d36]
nginx: worker process() [0x8052f61]
2016/01/19 20:11:55 [alert] 26475#0: worker process 26489 exited on signal 6
```
Pagespeed config: (inside server{} section)
```
pagespeed On;
pagespeed RewriteLevel CoreFilters;

# HTTPS Support
pagespeed FetchHttps enable,allow_self_signed;

pagespeed EnableFilters lazyload_images,collapse_whitespace,insert_dns_prefetch,dedup_inlined_images,defer_javascript,pedantic,trim_urls,sprite_images,extend_cache_pdfs,remove_comments,resize_mobile_images,inline_preview_images,insert_image_dimensions,convert_to_webp_lossless,local_storage_cache,inline_google_font_css,prioritize_critical_css,rewrite_style_attributes,move_css_to_head,move_css_above_scripts,outline_javascript,outline_css,combine_heads;

pagespeed FileCachePath /var/ngx_pagespeed_cache;

location ~ "\.pagespeed\.([a-z]\.)?[a-z]{2}\.[^.]{10}\.[^.]+" { add_header "" ""; }
location ~ "^/pagespeed_static/" { }
location ~ "^/ngx_pagespeed_beacon$" { }
```
nginx 1.8.0, CentOS 6.7

When i change to pagespeed Off; - nginx working.
```
# nginx -t
nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
[...]

PR #1098, opened 2016-01-21
Eliminate CHECK that fails (rarely) during the IPRO lookup.

oschaaf:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1096

### DIFF
--- a/src/ngx_base_fetch.cc
+++ b/src/ngx_base_fetch.cc
@@ -164,15 +164,13 @@ void NgxBaseFetch::ReadCallback(const ps_event_data& data) {
       return;
   }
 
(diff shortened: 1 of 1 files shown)


oschaaf:
Fix confirmed on the problem site with help from @grayfolk

jeffkaufman:
LGTM

jeffkaufman:
@oschaaf Should this be backported to 1.9?

jeffkaufman:
Looks like it doesn't need backporting: the code it's fixing isn't in 1.9.

oschaaf:
No this doesn't need to be backported.

Issue #1100, opened 2016-01-22
Valgrind complains about NgxBaseFetch::ReadCallback


oschaaf:
From https://groups.google.com/forum/#!msg/pagespeed-dev/u8pNl2fQL-Q/Vk_oGsrnCgAJ , https://github.com/pagespeed/ngx_pagespeed/pull/1098 introduced a valgrind error.

PR #1101, opened 2016-01-22
Fix valgrind complaint caused by NgxBaseFetch

oschaaf:
We should not use r->connection after we finalize the request.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1100


jeffkaufman:
LGTM

PR #1102, opened 2016-01-25
Adjust to library API change: no more spdy-specific config.

morlovich:
(Upcoming on mod_pagespeed end)


jeffkaufman:
LGTM

(Sorry I missed this earlier!)

PR #1103, opened 2016-01-27
Adjust our UA string to fix to make it compatible with fonts.g.c parsing on IE11.

morlovich:
Counterpart of:
https://github.com/pagespeed/mod_pagespeed/commit/f3639e84c0196a5f5151ff5e54ad57285db09b37

### DIFF
--- a/src/ngx_fetch.cc
+++ b/src/ngx_fetch.cc
@@ -938,8 +938,8 @@ void NgxFetch::FixUserAgent() {
     user_agent += "NgxNativeFetcher";
   }
   GoogleString version = StrCat(
-      " ", kModPagespeedSubrequestUserAgent,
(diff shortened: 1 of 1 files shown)


morlovich:
Note that I am having trouble running tests with native fetcher....

oschaaf:
Change LGTM

I can reproduce that testing with the native fetcher does not work at all, created https://github.com/pagespeed/ngx_pagespeed/issues/1104 for that

oschaaf:
FWIW, I worked on adding just enough ipv6 support to the native fetcher so we can run the tests with it again. State dumped over here:
https://github.com/pagespeed/ngx_pagespeed/compare/trunk-tracking...oschaaf-trunk-tracking-native-fetcher-ipv6?expand=1

Tests are passing again with it, barring the "Can inline Google Font API loader CSS" test, which probably needs this change to pass?

Issue #1064, opened 2015-12-09
Multiple Vary headers emitted


oschaaf:
On trunk-tracking, we somehow end up emitting three `Vary: Accept-Encoding` headers.

```bash
$ curl -I -X GET http://192.168.137.10:8090/mod_pagespeed_example/styles/index_style.css
HTTP/1.1 200 OK
Server: nginx/1.8.0
Content-Type: text/css
Connection: keep-alive
Vary: Accept-Encoding
ETag: "5666b969-2c7"
Accept-Ranges: bytes
Date: Wed, 09 Dec 2015 15:51:53 GMT
Expires: Wed, 09 Dec 2015 15:56:53 GMT
Cache-Control: max-age=300
X-Original-Content-Length: 711
Vary: Accept-Encoding
Vary: Accept-Encoding
Content-Length: 711

```

nginx.conf:

```
server {
    listen       8090;
    server_name  localhost;
    pagespeed on;
    pagespeed FileCachePath /tmp/ps;
    pagespeed RewriteLevel PassThrough;
    pagespeed CssPreserveUrls on;
    pagespeed EnableFilters prioritize_critical_css,debug,extend_cache;

    location / {
        root   /var/www/;
        index  index.html index.htm;
    }
}
```

deweydb:
Not sure if related, but i was having this issue when i had: gzip_vary on; in my http block, instead of the server block.

oschaaf:
@jeffkaufman is this worthy of being on the work-prioritization list?

jeffkaufman:
Added

HansVanEijsden:
Reminder, please.. would love to have this fixed too (request of one of my clients). 😃

Niwreg:
Same issue here. I'll follow the thread in case of updates :)

oschaaf:
Fix for this is in progress at https://github.com/pagespeed/ngx_pagespeed/tree/oschaaf-trunk-tracking-issue-1064, working on tests to prevent regression. 
I have yet to reproduce the situation again where we emit three `Vary: Accept-Encoding` headers which somehow seems hard now, but this fix gets rid of a reproducible double `Vary: Accept-Encoding` situation. 
If anyone is up for trying the patch, here it is: https://github.com/pagespeed/ngx_pagespeed/compare/trunk-tracking...oschaaf-trunk-tracking-issue-1064.patch

HansVanEijsden:
[...]

PR #1105, opened 2016-01-28
vary-header: Emit a single vary header in the IPRO flow

oschaaf:
The report from some time ago mentioned three Vary: headers,
but I can now only reproduce two using trunk-tracking plus the
original repro-configuration.

This fix unflags r->gzip_vary as set by the gzip module when PSOL
hands us Vary: Accept-Encoding, to make sure that nginx's core
header filter doesn't append another one.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1064


jeffkaufman:
LGTM

PR #1113, opened 2016-02-09
support centralized process changes

jeffkaufman:


oschaaf:
No comments, LGTM

src/ngx_rewrite_driver_factory.cc
```diff
@@ -250,4 +250,23 @@ void NgxRewriteDriverFactory::InitStats(Statistics* statistics) {
...
 
+void NgxRewriteDriverFactory::PrepareForkedProcess(const char* name) {
+  ngx_pid = ngx_getpid();  // Needed for logging to have the right PIDs.
+  SystemRewriteDriverFactory::PrepareForkedProcess(name);
+}
+
+void NgxRewriteDriverFactory::NameProcess(const char* name) {
+  SystemRewriteDriverFactory::NameProcess(name);
+
+  // Superclass set status with prctl.  Nginx has a helper function for setting
```
hillsp: I don't quite follow this. Why do you want to change the name *twice*?

test/nginx_system_test.sh
```diff
@@ -577,6 +577,18 @@ check test $(scrape_stat image_rewrite_total_original_bytes) -ge 10000
 # happens both before and after.
 start_test "Reload config"
 
+function check_process_names() {
+  if ! $USE_VALGRIND; then
+    # There should be one babysitter and controller running.  Under valgrind
+    # process labels are confused, so skip the check then.
+
+    check [ $(ps aux | grep -c 'nginx: pagespeed babysitte[r]') -eq 1 ]
```
hillsp: pgrep -c would be nice here, if it were installed.

jeffkaufman: Yes, but I don't want to add a pgrep dependency.

src/ngx_rewrite_driver_factory.cc
jeffkaufman: There are two kinds of process "names": there's the status, 16 chars you can set with `prctl`, and there's `argv[0]`.  Some ways of looking at processes use one, some use the other.  Nginx processes do typically set `argv[0]` so I'd like to do that to fit in.  That's awkward enough [1] that I only want to do it in situations where a helper function is available.

[1] https://cs.corp.google.com/piper///depot/google3/third_party/nginx/src/src/os/unix/ngx_setproctitle.c?q=ngx_setproctitle&l=79&dr=CSs

hillsp:
LGTM

jeffkaufman:
This went in as 8efb831b

PR #1115, opened 2016-02-10
 Support building ngx_pagespeed as a dynamic module

oschaaf:
As of 1.9.11, nginx supports loading dynamic modules
This change makes us support building ngx_pagespeed.so

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1116

### DIFF
--- a/config
+++ b/config
@@ -193,71 +193,122 @@ ngx_feature_test="
 # Test whether we have pagespeed and can compile and link against it.
 . "$ngx_addon_dir/cpp_feature"
 
-if [ $ngx_found = yes ]; then
-  ps_src="$ngx_addon_dir/src"
-  ngx_addon_name=ngx_pagespeed
-  NGX_ADDON_DEPS="$NGX_ADDON_DEPS \
-    $ps_src/log_message_handler.h \
-    $ps_src/ngx_base_fetch.h \
-    $ps_src/ngx_caching_headers.h \
-    $ps_src/ngx_event_connection.h \
-    $ps_src/ngx_fetch.h \
-    $ps_src/ngx_gzip_setter.h \
-    $ps_src/ngx_list_iterator.h \
-    $ps_src/ngx_message_handler.h \
-    $ps_src/ngx_pagespeed.h \
-    $ps_src/ngx_rewrite_driver_factory.h \
-    $ps_src/ngx_rewrite_options.h \
-    $ps_src/ngx_server_context.h \
-    $ps_src/ngx_url_async_fetcher.h \
-    $psol_binary"
-  NGX_ADDON_SRCS="$NGX_ADDON_SRCS \
-    $ps_src/log_message_handler.cc \
-    $ps_src/ngx_base_fetch.cc \
-    $ps_src/ngx_caching_headers.cc \
-    $ps_src/ngx_event_connection.cc \
-    $ps_src/ngx_fetch.cc \
-    $ps_src/ngx_gzip_setter.cc \
-    $ps_src/ngx_list_iterator.cc \
-    $ps_src/ngx_message_handler.cc \
-    $ps_src/ngx_pagespeed.cc \
-    $ps_src/ngx_rewrite_driver_factory.cc \
-    $ps_src/ngx_rewrite_options.cc \
-    $ps_src/ngx_server_context.cc \
-    $ps_src/ngx_url_async_fetcher.cc"
-  # Save our sources in a separate var since we may need it in config.make
-  PS_NGX_SRCS="$NGX_ADDON_SRCS"
-
-  if [ "$position_aux" = "true" ] ; then
-     HTTP_AUX_FILTER_MODULES="$HTTP_AUX_FILTER_MODULES $ngx_addon_name"
-  else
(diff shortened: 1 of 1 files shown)


PiotrSikora:
This looks very wrong to me, this patch:
1. always compilies both modules as static modules, even when compiling as a dynamic modules (using `--add-dynamic-module`), 
2. misuses `ngx_module_order`, which doesn't work with filters running after current filter, only earlier ones.

See: https://github.com/google/ngx_brotli/commit/86998c61fab726d4ee21fdf7e96f80e0cffbc71a

oschaaf:
@PiotrSikora Updated and force pushed. This will need to be rebased, but does this look better?

PiotrSikora:
@oschaaf: looks good (I still need to verify that `ngx_module_link` works with _two_ modules), but it seems that you also need to modify `config.make` in order to pass `--std=c++11` when compiling with `clang`, otherwise it doesn't build.

config
```diff
@@ -193,72 +193,116 @@ ngx_feature_test="
...
+                             | sed "s/$next/$next ngx_pagespeed/"`
+        # Make the etag header filter run immediately before range header filter.
+        HTTP_FILTER_MODULES=`echo $HTTP_FILTER_MODULES \
+                             | sed "s/ngx_pagespeed_etag_filter//" \
+                             | sed "s/ngx_http_range_header_filter_module/ngx_http_range_header_filter_module ngx_pagespeed_etag_filter/"`
+       echo "$HTTP_FILTER_MODULES"
     else
-      module=ngx_http_range_header_filter_module
+      # config.make is not executed for dynamic modules
+      CFLAGS="$CFLAGS -Wno-c++11-extensions"
```
oschaaf: @PiotrSikora config.make is not executed for dynamic modules, see https://github.com/nginx/nginx/blob/master/auto/make#L436

So this is the best I can come up with to allow us to build ngx_pagespeed.so with clang

oschaaf:
@PiotrSikora Comments addressed & force pushed

config
```diff
@@ -193,72 +193,116 @@ ngx_feature_test="
...
+                    ngx_http_gunzip_filter_module \
+                    ngx_http_userid_filter_module \
+                    ngx_http_headers_filter_module"
+
+    . auto/module
+
+    if [ $ngx_module_link != DYNAMIC ]; then
+        # ngx_module_order doesn't work with static modules,
+        # so we must re-order filters here.
+        HTTP_FILTER_MODULES=`echo $HTTP_FILTER_MODULES \
```
PiotrSikora: Style: this module uses `$(...)` for shell commands, so we should be consistent about it.

config
```diff
@@ -193,72 +193,116 @@ ngx_feature_test="
...
+                             | sed "s/ngx_pagespeed//" \
+                             | sed "s/$next/$next ngx_pagespeed/"`
+        # Make the etag header filter run immediately before range header filter.
+        HTTP_FILTER_MODULES=`echo $HTTP_FILTER_MODULES \
+                             | sed "s/ngx_pagespeed_etag_filter//" \
+                             | sed "s/ngx_http_range_header_filter_module/ngx_http_range_header_filter_module ngx_pagespeed_etag_filter/"`
+       echo "$HTTP_FILTER_MODULES"
```
PiotrSikora: This probably isn't needed anymore ;)

config
```diff
@@ -193,72 +193,116 @@ ngx_feature_test="
...
+                    ngx_http_sub_filter_module \
+                    ngx_http_addition_filter_module \
+                    ngx_http_gunzip_filter_module \
+                    ngx_http_userid_filter_module \
+                    ngx_http_headers_filter_module"
+
+    . auto/module
+
+    if [ $ngx_module_link != DYNAMIC ]; then
+        # ngx_module_order doesn't work with static modules,
```
PiotrSikora: Style: this module uses 2 spaces for intend, not 4.

config
```diff
@@ -193,72 +193,116 @@ ngx_feature_test="
...
+      CFLAGS="$CFLAGS -Wno-c++11-extensions"
     fi
-
-    HTTP_FILTER_MODULES=$(echo $HTTP_FILTER_MODULES |\
-      sed "s/$module/$module $ngx_addon_name/")
+else
+  CORE_LIBS="$CORE_LIBS $pagespeed_libs"
+  CORE_INCS="$CORE_INCS $pagespeed_include"
+  NGX_ADDON_SRCS="$PS_NGX_SRCS"
+  if [ "$position_aux" = "true" ] ; then
```
PiotrSikora: This feature is ignored in nginx-1.9.11+ code path, is that desirable behavior? Also, I'm not clear on what this is supposed to achieve in the first place...

PiotrSikora:
@oschaaf: thanks, LGTM (other than the stuff that I commented on, which can be mostly ignored).

oschaaf:
@PiotrSikora Thanks for reviewing! I updated to address your latest comments.

crowell:
@oschaaf is this ready to merge?

oschaaf:
@crowell yes

crowell:
LGTM, backported and tested on master.

Issue #1120, opened 2016-02-17
Old worker processes segfault on exit during reload


dotsam:
Running pagespeed 1.10.34.5 with nginx 1.8.1 on CentOS 7, I'm getting a segfault on reload (SIGHUP). Backtrace points to a problem in ps_exit_child_process, which looks to have been changed in d7f1c0dc4817f153da29ced58882cb0bc9787c2c

Running the same exact config with 1.9.32.10 and there is no issue. All debug info below.

uname -a
`Linux 3.10.0-327.4.5.el7.x86_64 #1 SMP Mon Jan 25 22:07:14 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux`

/etc/redhat-release
`CentOS Linux release 7.2.1511 (Core)`

Nginx version
```
nginx version: nginx/1.8.1
built with OpenSSL 1.0.1e-fips 11 Feb 2013
TLS SNI support enabled
[...]

PR #1122, opened 2016-02-17
Fix shutdown when ngx_pagespeed is completely disabled.

oschaaf:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1120


crowell:
lgtm

PR #1125, opened 2016-02-19
Allow varying on Accept header for ipro-for-browser.example.com

huibaolin:
For backward compatibility, only allow varying on Accept header for ipro-for-browser.example.com


oschaaf:
LGTM

Issue #1138, opened 2016-02-27
IPRO responses sometimes forward a bad cache control value


oschaaf:
Cloudflare sometimes shows an error page when serving an mp3 file. It does so when it receives a bad header that looks like: `Cache-Control: ?ax-age`, as seen in `error.log`.

The server is running ngx_pagespeed 1.10.33.5

oschaaf:
Initial report from ngx-pagespeed-discuss: https://groups.google.com/forum/#!topic/ngx-pagespeed-discuss/3cfPz35acRM

oschaaf:
The bad cache control header is emitted when we hit the flow where IPRO does not find a cache entry and subsequently starts recording the response. In this flow the original cache-control header is passed on which is owned by an `NgxBaseFetch` instance that will be released the passed on string is put on the wire. The fix is to copy the value, pull coming up.

jmarantz:
Thanks for tracking this down, Otto.  Is this related to specifying a threadpool in nginx?

oschaaf:
@jmarantz No this is unrelated to specifying a threadpool in nginx, though I was looking into both things at the same time because initially I though they might be related.

jeffkaufman:
Fixed with #1139 and 5264647a658712f2701cf19e3fb3d9269c262682

PR #1139, opened 2016-02-27
IPRO: Copy the cache control value to ensure a correct lifetime

oschaaf:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1138

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -492,6 +492,10 @@ ngx_int_t copy_response_headers_to_ngx(
 
   ngx_int_t i;
   for (i = 0 ; i < pagespeed_headers.NumAttributes() ; i++) {
+    // For IPRO cache misses, these gs_ variables may point to freed memory
+    // when nginx writes the headers to the output as the NgxBaseFetch instance
+    // that owns this memory gets released during request processing. So we
+    // copy these strings later on.
     const GoogleString& name_gs = pagespeed_headers.Name(i);
     const GoogleString& value_gs = pagespeed_headers.Value(i);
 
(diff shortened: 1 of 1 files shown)


src/ngx_pagespeed.cc
```diff
@@ -544,7 +544,8 @@ ngx_int_t copy_response_headers_to_ngx(
     }
 
     if (STR_EQ_LITERAL(name, "Cache-Control")) {
-      ps_set_cache_control(r, const_cast<char*>(value_gs.c_str()));
+      ps_set_cache_control(r, string_piece_to_pool_string(
+          r->pool, value_gs.c_str()));
```
jeffkaufman: I think you can just use `value_s` here; it looks like that's already available here for exactly this lifetime issue?

It also might be clearer to change the code above so that `value` is the output of `string_piece_to_pool_string` and then we can do away with `value_s`.

jeffkaufman: Another idea: what would you think of changing the `gs` suffix to `unsafe` or something to indicate that it's not a pool string?

oschaaf: @jeffkaufman Good ideas. I thought about this some ,what do you think about this?: We currently know when an NgxBaseFetch instance has detached from the nginx side (in which case we should not process further events received from psol for that NgxBaseFetch instance).
As such it should be ok to safely defer destruction of NgxBaseFetches. We could add an extra reference for this and decrement one more time from a separate request cleanup handler. If we do that, we don't have to copy strings, and we don't have unsafe values.

jeffkaufman: Ok, so the problem here is that we have some PSOL-allocated memory, and we need to keep that around until nginx is done with the request.  Two ways:

a) what your change does, and what the original code is trying to do: make a copy that's bound to the lifetime of the request, via a request pool.

b) what you're suggesting now: keep the PSOL memory around until we learn that the request has completed.

The advantage I'm seeing to (a) is that it's simpler and frees up memory (the rest of NgxBaseFetch) sooner, while the advantage of (b) is that it does a little less copying (the cache-control header value).

Saving copying <20 bytes doesn't seem worth rewriting this all?  So I'd lean towards (a) and just making the small fix.

oschaaf: @jeffkaufman updated to reflect a) and force pushed, PTAL?

jeffkaufman:
LGTM

PR #1145, opened 2016-03-07
Add system tests for image format, quality, and response in IPRO

huibaolin:
Add 2 servers for testing image rewrites in IPRO. Together with the existing server, ipro-for-browser.example.com, we will be testing testing image rewrites under different settings, in particular, include "AllowVaryOn" of "Auto", "Accept", and "None".


crowell:
LGTM! :+1:

Issue #1146, opened 2016-03-08
Possible regression: http2, nginx, chrome and windows


jdelgad:
Possible regression for https://github.com/pagespeed/ngx_pagespeed/issues/788

nginx 1.9.12, pagespeed 1.10.33.6-beta and the same configuration as posted here https://github.com/pagespeed/ngx_pagespeed/issues/1140

We are using SSL with a Let's Encrypt cert. Only Windows + Chrome users are seeing this issue. It ended up bringing the site to a stand still.

Here are a few errors we are getting in errors.log

2016/03/07 05:01:16 [error] 3692#0: *360332 upstream timed out (110: Connection timed out) while reading response header from upstream, client: XXX.XXX.XXX.XXX, server: www.XXXXXX.com, request: "GET /forums/ HTTP/2.0", upstream: "fastcgi://unix:/var/run/php-fpm/php-fpm.sock", host: "www.XXXXX.com"
2016/03/07 06:19:46 [error] 3691#0: *372356 connect() to unix:/var/run/php-fpm/php-fpm.sock failed (11: Resource temporarily unavailable) while connecting to upstream, client: XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX, server: www.XXXXX.com, request: "GET /path/to/file.jpg,g38273.pagespeed.ic.38c76x.jpg HTTP/2.0", upstream: "fastcgi://unix:/var/run/php-fpm/php-fpm.sock:", host: "www.XXXXX.com", referrer: "https://www.XXXXX.com/url/"
2016/03/07 15:59:53 [alert] 3688#0: *524522 http request count is zero while closing request, client: XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX, server: [::]:443
2016/03/07 16:31:45 [alert] 3690#0: *525244 open socket #45 left in connection 30
2016/03/07 17:53:23 [alert] 11173#0: *41609 http request count is zero while closing request, client: XXX.XXX.XXX.XXX, server: 0.0.0.0:443

jdelgad:
I had to turn off pagespeed and it appears to be working better. I can get logs. Don't have any core dump, but I don't think it caused any.

oschaaf:
@jdelgad If you could forward logs that may help figuring this out. Having logs files written with the log level set at debug would help even more.

jdelgad:
[...]

PR #1147, opened 2016-03-10
Fix for async waits vs client aborts and HTTP/2 RST_FRAMES

oschaaf:
This change makes nginx close the request when we receive a close
event or RST_STREAM while waiting for an async PSOL op to complete
by setting a read event handler, as inspired by the limit request
module from nginx.

Should fix https://github.com/pagespeed/ngx_pagespeed/issues/1146

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -1290,6 +1290,9 @@ ngx_int_t ps_decline_request(ngx_http_request_t* r) {
 
   r->count++;
   r->phase_handler++;
+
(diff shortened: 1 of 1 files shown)


HansVanEijsden:
Here the fix works great.
I applied it 5 days ago. Since then, no more "http request count is zero while closing request" messages in my Nginx error logs.

jdelgad:
After using this fix I agree with @HansVanEijsden as I'm no longer seeing "http request count is zero while closing request." I tried to use this branch, but couldn't get it to compile, so I had to do it manually, however.

jeffkaufman:
LGTM

Issue #1148, opened 2016-03-10
NgxFetch: ParseUrl() failed


creativeprogramming:
Hi,

I'm getting many errors like the following in the log file:

`2016/03/10 18:09:50 [error] 18594#18594: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
2016/03/10 18:10:00 [error] 18588#18588: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
2016/03/10 18:11:27 [error] 18594#18594: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
2016/03/10 18:11:28 [error] 18594#18594: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
2016/03/10 18:11:55 [error] 18587#18587: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
2016/03/10 18:12:00 [error] 18594#18594: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
2016/03/10 18:13:46 [error] 18593#18593: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
2016/03/10 18:13:57 [error] 18587#18587: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed
`

Is there a way to get more info about it? Can i configure the log level to print the URL that failed to be parsed? If i cannot configure it, also a code hack to make it printing is accepted as a suggestion. I'm trying to debug it

oschaaf:
@creativeprogramming See https://github.com/pagespeed/ngx_pagespeed/pull/1150

creativeprogramming:
Perfect thanks.

Here's what I'm getting now:

`2016/03/11 15:08:31 [error] 13827#13827: [ngx_pagespeed 1.10.33.6-0] NgxFetch: ParseUrl() failed for [https://it/articles/7917-article-sef-url]:host not found`

the url should be: https:/mysite.it/it/articles/7917-article-sef-url, In other words the hostname is stripped from the url.

Can it be due to the fact I'm using **protocol relative URIs** to link resources? e.g. in my html i've 
 `<base href="//mysite.it/it..." `?

[...]

PR #1150, opened 2016-03-11
NgxFetch: Log url and error when url parsing fails.

oschaaf:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1148


jmarantz:
lgtm

Issue #1149, opened 2016-03-11
Set Cache-Control: public on pagespeed optimized resources


JoyceBabu:
Currently pagespeed adds `Cache-Control:max-age=31536000` to optimized resources. Is it possible to add `public` to it?

Google CDN (alpha)  requires 'Cache-Control: public' to be present for resources that must be cached.

jmarantz:
Hi -- I'm getting some clarification from some people that know more about
the CDN.

jmarantz:
Adding "public" to the cache-control value is a requirement for Google CDN, but is not required by the HTTP spec.   https://cloud.google.com/compute/docs/load-balancing/http/cdn describes the Google requirements, which are pretty clear.

To meet these without adding an unnecessary qualifier for other caches, we can gate this logic on the presence of response-header
  Via: 1.1 google

jmarantz:
Joyce: is it possible for you to add a workaround for this problem in your nginx config?  That might help you get better behavior until we can add this functionality to a PageSpeed release.

JoyceBabu:
I tried disabling all rewriting rules, and using `add_header 'Cache-Control' public;`. This resulted in two Cache-Control headers
```
Cache-Control:max-age=604800
Cache-Control: public
```

Google CDN did not cache those resources, even though it is technically equivalent to `Cache-Control:max-age=604800, public`

> To meet these without adding an unnecessary qualifier for other caches, we can gate this logic on the presence of response-header

Did you mean **request**-header?

> Joyce: is it possible for you to add a workaround for this problem in your nginx config? That might help you get better behavior until we can add this functionality to a PageSpeed release.

I am unable to modify the Cache-Control headers from nginx config. I tried using `POSITION_AUX=true`, still I am unable to modify the headers. I remember seeing a comment by Jeff in another issue that nginx `location` blocks don't when LoadFromFile is enabled.

jmarantz:
Yes I meant the Via request-header :)

there is not an nginx equivalent to apache's "header set" directive?

JoyceBabu:
> there is not an nginx equivalent to apache's "header set" directive?

Yes, there is the `add_header` directive and the `more_set_headers` directive. But from what I understand, the `header` filter appears before `ngx_pagespeed` in the nginx filter chain. Hence, headers are always intercepted and modified by `ngx_pagespeed`. 

[...]

PR #1155, opened 2016-03-12
Remove ngx_pagespeed from HTTP_FILTER_MODULES when POSITION_AUX=true

### LINKED ISSUE #1154: POSITION_AUX=true broken in nginx 1.9.11+
Setting `POSITION_AUX=true` is not working in nginx 1.9.11+. The module `ngx_pagespeed` is appearing in both `HTTP_FILTER_MODULES` and `HTTP_AUX_FILTER_MODULES `.  `ngx_pagespeed` is appearing by default in `HTTP_FILTER_MODULES`. When `POSITION_AUX=true`, the default entry needs to be removed before it is added to `HTTP_AUX_FILTER_MODULES`.

JoyceBabu:
Remove ngx_pagespeed from HTTP_FILTER_MODULES when POSITION_AUX=true. Fixes #1154


JoyceBabu:
I have signed the CLA.

oschaaf:
LGTM

jmarantz:
What's the story with this CL?  Joyce: is this something you want commit?

oschaaf:
Landing this would require a rewrite, because things have changed a lot in the code that this touches. @centminmod do you use this? @JoyceBabu do you? 

`POSITION_AUX` is an undocumented feature that is only rarely needed afaict -- so I don't think it is used a lot. If no-one here is using it, maybe we should just close this?

JoyceBabu:
@jmarantz I am sorry, I don't recall why I had to use the `POSITION_AUX` feature. I believe it was due to https://github.com/pagespeed/ngx_pagespeed/issues/1149#issuecomment-195794246 .

@oschaaf I am now using Apache + ModPagespeed, hence the feature is not used by me. But considering the issue mentioned in the linked comment, I think there might still be valid use cases for the feature.

oschaaf:
Closing this PR. There might be a use-case for it, but it seems rare and I'm not sure it currently is worth the effort of rewriting this to work with the current code base. Let's see if https://github.com/pagespeed/ngx_pagespeed/issues/1149#issuecomment-195794246 pops up more, if it gets traction we can revisit this.

PR #1156, opened 2016-03-15
controller: don't fail tests when running in emacs

jeffkaufman:
Apparently, when running in emacs and piped into another command, ps limits
it's output to COLUMNS making check_process_names fail.  Telling ps to use
wide, wide output with 'ww' fixes this.

Additionally, use check_from so that if this test does fail we can debug
it better, seeing what grep was looking at.


jmarantz:
lgtm

PR #1157, opened 2016-03-16
Turn on C++11 for all compilers, not just clang.

morlovich:
Explain why we do it via config.make better.
Remove partial workaround for config.make not invoked for dynamic modules,
nginx dev fixes the root issue, and the workaround is insufficient anyway.


oschaaf:
LGTM

PR #1161, opened 2016-03-22
ignore another serf warning that appears from time to time

jmarantz:


oschaaf:
@jmarantz I'm wondering if this message is something that should be ignored, do you know if this was for a loopback a fetch from an external domain?

jmarantz:
Actually I put this in my script some weeks back to make it through our
standard checkin tests.  It seems to happen only sometimes.  The checkin
tests had passed except for the warning, and our usual fetch_until logic in
the scripts got past whatever was failing.

However if you think this is a transient error that might not occur any
longer, I can always revert it and put it back at a later time if I see it.

oschaaf:
@jmarantz would it be possible to forward me nginx's error.log files when this happens before merging this?

PR #1171, opened 2016-04-06
Companion to mps change to set s-maxage on unoptimized ipro resources

jeffkaufman:
mps change: https://github.com/pagespeed/mod_pagespeed/commit/689e3f1a0dd70c8585a2304c044b4408a8da0bc3

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -422,6 +422,32 @@ ngx_int_t ps_set_cache_control(ngx_http_request_t* r, char* cache_control) {
   return NGX_OK;
 }
 
+// Returns false if the header wasn't found.  Otherwise sets cache_control and
+// returns true;
+bool ps_get_cache_control(ngx_http_request_t* r, GoogleString* cache_control) {
+  // Use headers_out.cache_control instead of looking for Cache-Control in
+  // headers_out.headers, because if an upstream sent multiple Cache-Control
+  // headers they're already combined in headers_out.cache_control.
+  auto ccp = static_cast<ngx_table_elt_t**>(r->headers_out.cache_control.elts);
+  if (ccp == nullptr) {
+    return false;  // Header not present.
+  }
+  bool first_segment = true;
+  for (ngx_uint_t i = 0; i < r->headers_out.cache_control.nelts; i++) {
+    if (ccp[i]->hash == 0) {
+      continue;  // Elements with a hash of 0 are marked as excluded.
+    }
+    if (first_segment) {
+      first_segment = false;
+    } else {
+      cache_control->append(", ");
+    }
+    cache_control->append(reinterpret_cast<char*>(ccp[i]->value.data),
+                          ccp[i]->value.len);
+  }
+  return true;
+}
+
 template<class Headers>
 void copy_headers_from_table(const ngx_list_t &from, Headers* to) {
   // Standard nginx idiom for iterating over a list.  See ngx_list.h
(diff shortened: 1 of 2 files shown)


src/ngx_pagespeed.cc
```diff
@@ -422,6 +422,20 @@ ngx_int_t ps_set_cache_control(ngx_http_request_t* r, char* cache_control) {
   return NGX_OK;
 }
 
+StringPiece ps_get_cache_control(ngx_http_request_t* r) {
+  ngx_table_elt_t* header;
+  NgxListIterator it(&(r->headers_out.headers.part));
+  while ((header = it.Next()) != NULL) {
+    if (STR_CASE_EQ_LITERAL(header->key, "Cache-Control")) {
```
oschaaf: I'm wondering if we should also take the `cache_control` field of `ngx_http_headers_out_t` [1] into account?

http://lxr.nginx.org/source/src/http/ngx_http_request.h#0271

jeffkaufman: My understanding is that when headers are written initially they're both added as a header and parsed into `ngx_http_headers_out_t.cache_control`, and after that point anyone who modifies headers is supposed to change both places.  For example, changing both places is what our `ps_set_cache_control` above does, and what `ngx_http_add_cache_control` does. So the literal header is what's sent out, and the parsed version is for the convenience of anyone who wants to read the current headers without parsing them anew.

So it's safe for us to read either `r->headers_out.headers` and find the `Cache-Control` one, or to reassemble a string from `r->headers_out.cache_control`.

@PiotrSikora , does this comment sound right to you?

PiotrSikora: You should really use `r->headers_out.cache_control`, otherwise you're going to miss some directives in case upstream sent multiple `Cache-Control` headers, whereas NGINX is going to send all `Cache-Control` headers downstream and (if configured) it will log combined value, covering all `Cache-Control` headers.

jeffkaufman: Thanks!  Switched this to using `ngx_http_headers_out_t.cache_control`.

oschaaf:
LGTM

src/ngx_pagespeed.cc
```diff
@@ -422,6 +422,32 @@ ngx_int_t ps_set_cache_control(ngx_http_request_t* r, char* cache_control) {
...
 }
 
+// Returns false if the header wasn't found.  Otherwise sets cache_control and
+// returns true;
+bool ps_get_cache_control(ngx_http_request_t* r, GoogleString* cache_control) {
+  // Use headers_out.cache_control instead of looking for Cache-Control in
+  // headers_out.headers, because if an upstream sent multiple Cache-Control
+  // headers they're already combined in headers_out.cache_control.
+  auto ccp = static_cast<ngx_table_elt_t**>(r->headers_out.cache_control.elts);
+  if (ccp == NULL) {
```
hillsp: nullptr if you're going to go all C++11

src/ngx_pagespeed.cc
```diff
@@ -422,6 +422,32 @@ ngx_int_t ps_set_cache_control(ngx_http_request_t* r, char* cache_control) {
...
+    return false;  // Header not present.
+  }
+  bool first_segment = true;
+  for (unsigned int i = 0; i < r->headers_out.cache_control.nelts; i++) {
```
crowell: `ngx_uint_t` instead of `unsigned int`

src/ngx_pagespeed.cc
```diff
@@ -422,6 +422,32 @@ ngx_int_t ps_set_cache_control(ngx_http_request_t* r, char* cache_control) {
...
+    if (ccp[i]->hash == 0) {
+      continue;  // Elements with a hash of 0 are marked as excluded.
+    }
+    if (first_segment) {
+      first_segment = false;
+    } else {
+      cache_control->append(", ");
+    }
+    cache_control->append(reinterpret_cast<const char*>(ccp[i]->value.data),
```
crowell: should this be a `const_cast` instead of a `reinterpret_cast`? the data field is a u_char*, so we're adding constness here.

jeffkaufman: removed the unnecessary const

(we need the reinterpret_cast because we're converting nginx's unsigned character type to c++'s signed one)

src/ngx_pagespeed.cc
jeffkaufman: Done

src/ngx_pagespeed.cc
jeffkaufman: Done

crowell:
lgtm :+1:

PR #1176, opened 2016-04-11
fix up the description in the run_tests.sh

crowell:
run_tests had the wrong invocation of the script in the comments. fix
that.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -23,9 +23,11 @@
 # Exits with status 2 if command line args are wrong.
 #
 # Usage:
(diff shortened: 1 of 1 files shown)


crowell:
this also belongs in trunk-tracking once this is in here.

oschaaf:
`PAGESPEED_TEST_HOST` can be passed in as the argument after the nginx binary though. Should we still doc that? 

```
oschaaf@ubuntu:~/code/google/nps-issue-1175/test$ grep -ir PAGESPEED_TEST_HOST
...
run_tests.sh:PAGESPEED_TEST_HOST="$5"
...
```

crowell:
fixed and squashed.

oschaaf:
lgtm!

Issue #1175, opened 2016-04-10
Pagespeed breaks form submission?


amourinho:
Even with simple code like this, pagespeed seems to break form submission.

<?
if(isset($_POST['submit'])){
        $name = $_POST['inName'];
}
echo "Hello " . $name;
?>
`<form method="post">
<input type="text" id="inName" name="inName"><button type="submit" id="submit" name="submit">Submit</button>
</form>
`

Does not work once pagespeed is enabled. Works properly when pagespeed disabled.
Tested on pagespeed versions: 1.10.33.6 and 1.11.33.0
Nginx version: 1.9.14

No error's are thrown by browser, or nginx. The page just times out. Any ideas on how to go about fixing this issue?

jmarantz:
Does it help to disable pagespeed on the form's submit urls via "pagespeed
disallow URL_WILDCARD:" ?

oschaaf:
I'm wondering if you can reproduce this on nginx `1.9.13`? Nginx `1.9.14` seems to contain some changes in the code that processes the request body (which is relevant for form submissions).

amourinho:
I switched to Nginx 1.9.13, seems to have fixed the issue. So there's some compatibility issue with their latest. Thanks for the ideas and help guys.

nikolay:
@oschaaf I wish Nginx truly followed SemVer! Patch versions more often than not are not backward compatible!

oschaaf:
I'm still  looking into this. I just noticed that with nginx 1.9.14, performing a POST via  HTTP/2 to a non-existing url will also result in a protocol violation, even when build without ngx_pagespeed. 
After some debugging I figured that out, as I was trying to make us not handle POST requests to `.pagespeed.` resources amongst others.

oschaaf:
The limit request module seems to suffer as well, and uses a similar flow to IPRO.
Posted to nginx-dev: https://forum.nginx.org/read.php?29,266134

oschaaf:
Update:
- An issue was filed at Chrome for this: https://bugs.chromium.org/p/chromium/issues/detail?id=603182
- NGINX added a workaround https://github.com/nginx/nginx/commit/7691b9750e0d3a2137e9d9188db65f7e39624b97

Angristan:
Any update ?

jmarantz:
The bug in Chrome has been assigned, but doesn't appear to have been updated for more than a month.

[...]

PR #1177, opened 2016-04-11
Fix request processing stalling on H/2 POSTS (as of nginx 1.9.14)

oschaaf:
Nginx 1.9.14 introduces changes that seem to make us stall request
processing when we receive a POST request on HTTP/2. Debugging shows
we initiate the IPRO lookup, resume processing and things do not
proceed after that.

I suspect this is because NGINX now expects us to initiate reading of
the request body before we embark on an async wait operation in the
preaccess phase, but I yet have to prove that.

This change makes us only accept doing work for POST requests for
beacons in the preaccess phase. This path actually read the request
body and passes manual tests. I think is sufficient to tackle the
problem - and it looks like a change we should make anyway.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1175

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -1786,11 +1786,13 @@ ngx_int_t ps_resource_handler(ngx_http_request_t* r,
 
   CHECK(!(html_rewrite && (ctx == NULL || ctx->html_rewrite == false)));
 
-  if (!html_rewrite &&
+  // TODO(oschaaf): Cover HTTP/2 in the system tests, especially IPRO and POSTS.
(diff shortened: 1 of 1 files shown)


crowell:
lgtm

oschaaf:
note - this change would be good to have on master as well if it passes review. I see I made a logical error in the condition, I'll force push the correct change shortly - sorry!

crowell:
LGTM now :+1:

oschaaf:
yeah but somehow we're not optimizing html responses to post requests with this change, still looking into that. I'll ping this issue when I figure that out

Issue #1192, opened 2016-05-09
cache-purging does not work for in-place resource optimization


deweydb:
I've tried both via ssh and also via /pagespeed_admin/cache#purge_cache
When i click the "purge entire cache" button it says "Purge successful" I can see that the timestamp on the cache.purge file has changed, but all old files still persist.

Could this be permissions related?  Is there some place that failures to delete files from the /v3/ folder are logged?

I experience the same behaviour when trying to purge individual files as well.

Any help greatly appreciated!

Configs:
server:
https://p.ngx.cc/317bf88f36c6648a
host:
https://p.ngx.cc/8b9605488d0773f9

oschaaf:
Purging the cache does not physically remove the current pagespeed-cache files from disk (but all the the entries that exist at the time of purging will be considered stale). 

Assuming that answers your question, I'm closing this - but please re-open if you actually have a problem with stale content being served!

deweydb:
Hi Otto,

Yes, stale content continues to be served.  The only way i have been able to get it to stop serving stale content is to literally delete my entire cache folder.  i.e. the folder defined at:  "pagespeed FileCachePath /var/ngx_pagespeed_cache;"

Something like:
`service nginx stop
rm -Rf /var/ngx_pagespeed_cache
service nginx start`

oschaaf:
Looking at the configuration you posted, you are also using downstream caching.
Did you purge the proxy cache as well?

deweydb:
Yes.

deweydb:
So, if i understand this correctly, the intended functionality is that stale cache items remain on disk even after cache purge.  I presume nginx page speed module does some sort of check to see if this item is stale, before re-serving the file from disk, and instead generating a new file.  What is this check? is it based on dates?  Perhaps nginx_page_speed module is operating on a different clock setting than my system?  

When a stale cache file is detected, and requested again from the browser, what is the intended behaviour? is this file deleted and re-created? or is an entirely new file created, and the old file remains?  Is there any way to see logs of these actions, delete, create, overwrite actions that occur in the cache folder?  I believe this will lend a lot of insight into what the root of the problem is.

deweydb:
An example can be seen here:
https://www.tivoresearch.com/contact/
Please note the CSS rule that is making the blue text in the black box have an underline is:
#map-info-box .address-body a{
	text-decoration: underline;
}
This css is actually commented out, as can be see by requesting the document with a pagepseed bypass:
[...]

Issue #1172, opened 2016-04-06
Purging the cache causes a non-rewritten, old version of a file to be served


AssortedParrot:
I'm using nginx 1.8.1 with ngx_pagespeed 1.10.33.2 on Debian 8.4, running Linux 3.16 on a x64 VM.

I'm having a problem where after purging the cache for a file, any subsequent requests will return a old version of the file without any filters applied.

```
root@mybox:/var/www/htdocs# cat scripts.js
// Some comment
// Another comment
stuff.that('looks', {"like": 'a script'});
alert('some alert');
root@mybox:/var/www/htdocs# cat request.txt
GET /scripts.js HTTP/1.1
Host: mysite.com
```

Not sure if it's related to this issue, but if I make the request three times, I get three different responses; after the third request the response seems to stay more or less the same. Starting with the third request, the file is rewritten, so clearly ngx_pagespeed is active, but I'm intrigued by the missing X-Page-Speed header.

```
root@mybox:/var/www/htdocs# nc localhost 80 < request.txt
HTTP/1.1 200 OK
Server: nginx/1.8.1
Date: Wed, 06 Apr 2016 16:12:36 GMT
Content-Type: application/javascript
Content-Length: 99
Last-Modified: Tue, 05 Apr 2016 15:01:11 GMT
Connection: keep-alive
ETag: "5703d337-68"
Accept-Ranges: bytes

// Some comment
// Another comment
stuff.that('looks', {"like": 'a script'});
alert('some alert');^C
root@mybox:/var/www/htdocs# nc localhost 80 < request.txt
HTTP/1.1 200 OK
Server: nginx/1.8.1
Content-Type: application/javascript
Connection: keep-alive
ETag: "5703d337-68"
Accept-Ranges: bytes
Date: Wed, 06 Apr 2016 16:12:36 GMT
Expires: Wed, 06 Apr 2016 16:17:36 GMT
Cache-Control: max-age=300
X-Original-Content-Length: 99
Vary: Accept-Encoding
Vary: Accept-Encoding
Content-Length: 99

// Some comment
// Another comment
stuff.that('looks', {"like": 'a script'});
alert('some alert');^C
root@mybox:/var/www/htdocs# nc localhost 80 < request.txt
HTTP/1.1 200 OK
Server: nginx/1.8.1
Content-Type: application/javascript
Connection: keep-alive
Accept-Ranges: bytes
X-Original-Content-Length: 60
Vary: Accept-Encoding
[...]

PR #1193, opened 2016-05-10
respect purge requests when serving ipro requests from ngx_pagespeed

jmarantz:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1192

May also fix https://github.com/pagespeed/ngx_pagespeed/issues/1172

### DIFF
--- a/src/ngx_base_fetch.cc
+++ b/src/ngx_base_fetch.cc
@@ -23,6 +23,8 @@
 
 #include "ngx_pagespeed.h"
 
+#include "net/instaweb/rewriter/public/rewrite_driver.h"
+#include "net/instaweb/rewriter/public/rewrite_options.h"
 #include "net/instaweb/rewriter/public/rewrite_stats.h"
 #include "pagespeed/kernel/base/google_message_handler.h"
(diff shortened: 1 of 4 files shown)


oschaaf:
LGTM

hillsp:
It make sense to move some of this Fetch stuff into a shared class (in system/)? The specific problem here was that Apache and nginx had diverged in a subtle way, whereas IsCachedResultValid isn't actually specific to either.

jmarantz:
Yes -- sharing a common base class between NgxFetch and ApacheFetch makes a lot of sense.  Adding a TODO to (a) avoid delaying this fix and (b) that might be harder to backport due to massive changes in ApacheFetch that are not in 1.11.

hillsp:
SGTM. It'd be nice to see the travis check pass, but I don't see anything on the UI for retry (I may lack bits).

jmarantz:
the test I haven't committed yet proves this fix works.  Unfortunately to check in the test I need to fix some infrastructural pieces of nginx and it's way easier to do that if I submit this first.

PR #1196, opened 2016-05-11
Fix IPRO and missing Content-Length headers in 32-bit builds.

hillsp:


crowell:
lgtm

src/ngx_base_fetch.cc
```diff
@@ -255,6 +256,13 @@ ngx_int_t NgxBaseFetch::CollectAccumulatedWrites(ngx_chain_t** link_ptr) {
 }
 
 ngx_int_t NgxBaseFetch::CollectHeaders(ngx_http_headers_out_t* headers_out) {
+  // nginx defines _FILE_OFFSET_BITS to 64, which changes the size of off_t.
+  // If a standard header is accidentally included before the nginx header,
+  // on a 32-bit system off_t will be 4 bytes and we don't assign all the
+  // bits of content_length_n. Sanity check that did not happen.
+  // This could use static_assert, but this file is not built with --std=c++11.
+  bool sanity_check_off_t[sizeof(off_t) == 8 ? 1 : -1] __attribute__ ((unused));
```
jmarantz: nice

src/ngx_pagespeed.cc
```diff
@@ -492,7 +492,8 @@ void copy_response_headers_from_ngx(const ngx_http_request_t* r,
 
   // When we don't have a date header, set one with the current time.
   if (headers->Lookup1(HttpAttributes::kDate) == NULL) {
-    headers->SetDate(ngx_current_msec);
```
jmarantz: should we comment the hell out of this variable and its ancestors that it is totally bogus on 32-it?

hillsp: I did check that we're not using it anywhere else. I don't think a comment would be much use, though something to actually hide/break the variable wouldn't be an awful idea. Not sure exactly how to go about that, though.

jmarantz: Never mind; I see this is just a big fat global bomb supplied by nginx that we are using.

jmarantz:
lgtm

PR #1197, opened 2016-05-11
During system-tests, don't write into the source-controlled htdocs area.

jmarantz:
Instead, use a new (not-yet-committed) target in Makefile.tests to build
a mirror htdocs in test/tmp/root that sym-links the large readonly
section and deep copies the test areas for purge and flush tests.

This must be committed alongside https://github.com/pagespeed/mod_pagespeed/pull/1306


crowell:
lgtm

oschaaf:
LGTM

PR #1200, opened 2016-05-14
Pass the install-path into Makefile

jmarantz:
... rather than having the Makefile depend on PWD.

### DIFF
--- a/test/nginx_system_test.sh
+++ b/test/nginx_system_test.sh
@@ -61,8 +61,9 @@ rm -rf "$SERVER_ROOT"
 mkdir -p "$SERVER_ROOT"
 export APACHE_DOC_ROOT="$SERVER_ROOT"
 
-mkdir -p $APACHE_DOC_ROOT
-cd "$MOD_PAGESPEED_DIR/src/install/" && make -f Makefile.tests setup_doc_root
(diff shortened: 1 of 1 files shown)


test/nginx_system_test.sh
```diff
@@ -61,8 +61,9 @@ rm -rf "$SERVER_ROOT"
 mkdir -p "$SERVER_ROOT"
 export APACHE_DOC_ROOT="$SERVER_ROOT"
 
-mkdir -p $APACHE_DOC_ROOT
-cd "$MOD_PAGESPEED_DIR/src/install/" && make -f Makefile.tests setup_doc_root
+mkdir -p "$APACHE_DOC_ROOT"
+make -f "$APACHE_DOC_SRC/Makefile.tests" setup_doc_root \
+  "INSTALL_DATA_DIR=$APACHE_DOC_SRC"
```
crowell: should be `INSTALL_DATA_DIR="$APACHE_DOC_SRC"` (location of quotation mark). and switch the variable definition to before the target.

`make -f "$APACHE_DOC_SRC/Makefile.tests" INSTALL_DATA_DIR="$APACHE_DOC_SRC" setup_doc_root`

crowell:
one small nit to make it consistent with other makefiles. LGTM modulo that.

PR #1205, opened 2016-05-26
Update for API: kProcessScope is about to become kLegacyProcessScope

morlovich:
The renaming is to make it clear that kProcessScopeStrict (aka what
ngx_pagespeed does for everything already) is what new code should be
using, rather than current kProcessScope/upcoming kLegacyProcessScope
which does weird backwards compatible things in mod_pagespeed.


morlovich:
Note that this shouldn't build yet, since corresponding PSOL change isn't in yet. Don't want to submit it w/o reviewed nginx counterpart.

oschaaf:
LGTM

PR #1210, opened 2016-06-08
Suppress controller warnings in test, fix up kill signals

hillsp:


test/nginx_system_test.sh
```diff
@@ -1315,7 +1315,7 @@ start_test Shutting down.
 fire_ab_load
 
 if $USE_VALGRIND; then
-    kill -s quit $VALGRIND_PID
+    kill -s TERM $VALGRIND_PID
```
oschaaf: I think this switches us to a quick shutdown instead of a graceful one when testing a stressed shutdown under Valgrind. Is that intentional?

hillsp: It was not, plus it causes an additional log message that is problematic for the system test. Fixed.

oschaaf:
Lgtm

test/nginx_system_test.sh
```diff
@@ -240,7 +240,7 @@ if $USE_VALGRIND; then
             --show-possibly-lost=no --log-file=$TEST_TMP/valgrind.log \
             --suppressions="$this_dir/valgrind.sup" \
       $NGINX_EXECUTABLE -c $PAGESPEED_CONF) & VALGRIND_PID=$!
-  trap "echo 'terminating valgrind!' && kill -s sigterm $VALGRIND_PID" EXIT
+  trap "echo 'terminating valgrind!' && kill -s TERM $VALGRIND_PID" EXIT
```
morlovich: What's the behavior different here?

test/nginx_system_test.sh
```diff
@@ -1356,6 +1357,8 @@ OUT=$(cat "$ERROR_LOG" \
     | grep -v "\\[warn\\].*Resource based on.*ngx_pagespeed_statistics.*" \
     | grep -v "\\[warn\\].*Canceling 1 functions on sequence Shutdown.*" \
     | grep -v "\\[warn\\].*using uninitialized.*" \
+    | grep -v "\\[warn\\].*Controller process .* exited with wait status 9" \
+    | grep -v "\\[warn\\].*Controller process .* exited with wait status 15" \
```
morlovich: So is this the one functional bit, the rest being hygienic?

hillsp: Exactly.

test/nginx_system_test.sh
hillsp: It's not, I just normalised onto one spelling of the signals.

morlovich:
LGTM

PR #1217, opened 2016-07-04
Make ngx_pagespeed behave like mod_pagespeed with regard to flushing

oschaaf:
This change makes ngx_pagespeed listen to the FollowFlushes option.
When set to on (=default), ngx_pagespeed will forward incoming flushes
to ProxyFetch. When writing output, we'll now also set the flush flag on
the buffers we are about to send downstream.

Companion to mps PR https://github.com/pagespeed/mod_pagespeed/pull/1066


morlovich:
LGTM pending the PSOL change...

morlovich:
(Confirm the combination passes tests)

PR #1218, opened 2016-07-04
Fix for when there is only server-level configuration

oschaaf:
When there is no top-level configuration but only server configuration,
we don't have main options to set and consider. This prevents
a NULL dereference in that case. (Later on we'll write out a message
about why we cannot startup with this configuration).

### DIFF
--- a/src/ngx_rewrite_driver_factory.cc
+++ b/src/ngx_rewrite_driver_factory.cc
@@ -210,7 +210,9 @@ void NgxRewriteDriverFactory::StartThreads() {
 void NgxRewriteDriverFactory::SetMainConf(NgxRewriteOptions* main_options) {
   // Propagate process-scope options from the copy we had during nginx option
   // parsing to our own.
-  default_options()->MergeOnlyProcessScopeOptions(*main_options);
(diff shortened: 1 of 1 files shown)


jeffkaufman:
LGTM

jeffkaufman:
Can you file a bug for this, so we can reference that bug when we release the bugfix release?

morlovich:
This looks potentially only broken by my fix to the max image rewrites
regression, so it might not actually be in any release
(unless that change is in latest 1.11?)

jeffkaufman:
> my fix to the max image rewrites regression

Link to bug or commit?

(If that fix is going out with the 1.11 bugfix then this patch needs to go out with it too.)

morlovich:
Commit would be https://github.com/pagespeed/ngx_pagespeed/commit/15c6f7c345f9c155b5b9097bd85642d3406a81f4
The mod_pagespeed portion is in there already, added this + dependency.

Updated the backport candidate spreadsheet --- I think you can remove your separate row on this...

jeffkaufman:
row removed

PR #1225, opened 2016-07-15
Check options->IsAllowed() before rewriting html.

oschaaf:
While diagnosing a problem on the mps discussion [1] list I noticed that when using the following configuration ngx_pagespeed would still pass html through PageSpeed:

```
        pagespeed Disallow *;
```

This PR changes that, which seems on par with mod_pagespeed when I look at the code.

[1] https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/mod-pagespeed-discuss/ReOcIWbH7DE/RgQSIhUyBgAJ

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -2004,7 +2004,7 @@ ngx_int_t ps_resource_handler(ngx_http_request_t* r,
 
   }
 
-  if (html_rewrite) {
+  if (html_rewrite && options->IsAllowed(url.Spec())) {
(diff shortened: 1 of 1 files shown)


jeffkaufman:
LGTM

oschaaf:
@jeffkaufman do I need to create an issue for this before merging it?

jeffkaufman:
Filing a bug would be good so (a) you can label this pull request as fixing that bug and (b) we can list the bug in our release notes.

Issue #1190, opened 2016-05-05
adding ?PageSpeedFilters=+debug to URL enables other filters


bessey:
Steps to repro:

1. Add `pagespeed RewriteLevel PassThrough;` to nginx config.
2. Navigate to a URL and add `?PageSpeedFilters=+debug` to it.

Observe that PageSpeed outputs:

``` html
<!-- 
mod_pagespeed on
Filters:
ah	Add Head
cc	Combine Css
jc	Combine Javascript
gp	Convert Gif to Png
jp	Convert Jpeg to Progressive
jw	Convert Jpeg To Webp
mc	Convert Meta Tags
pj	Convert Png to Jpeg
ws	When converting images to WebP, prefer lossless conversions
db	Debug
ei	Cache Extend Images
fc	Fallback Rewrite Css 
if	Flatten CSS Imports
hw	Flushes html
ci	Inline Css
ii	Inline Images
il	Inline @import to Link
ji	Inline Javascript
js	Jpeg Subsampling
rj	Recompress Jpeg
rp	Recompress Png
rw	Recompress Webp
ri	Resize Images
cf	Rewrite Css
jm	Rewrite External Javascript
jj	Rewrite Inline Javascript
cu	Rewrite Style Attributes With Url
cp	Strip Image Color Profiles
md	Strip Image Meta Data
...
-->
```

If i refresh the page a couple times these extra filters start kicking in too (e.g. URLs start getting rewritten even though I have `pagespeed CssPreserveURLs On;` in my nginx config.

oschaaf:
Could you post the full pagespeed-related configuration?

bessey:
Sure:
```
## GLOBAL PAGESPEED CONFIGURATION
pagespeed FileCachePath                   /home/nps-cache;
pagespeed CreateSharedMemoryMetadataCache /home/nps-cache 66000;
pagespeed MessageBufferSize     1000000;
pagespeed FileCacheSizeKb       2048000;
pagespeed LRUCacheKbPerProcess  16384;
pagespeed LRUCacheByteLimit     32768;

## ADMIN
pagespeed Statistics on;
pagespeed StatisticsLogging on;
pagespeed GlobalAdminPath /pagespeed_admin;


server {
  listen    443 ssl http2;
  server_name  www.example.com;

  #### PAGESPEED CONFIGURATION ####
  pagespeed on;
  # I must always be on, even if no experiments are running, or header based filter modification will not work
[...]

PR #1228, opened 2016-07-22
rewrite-options: don't turn on CoreFilters just because of query params

jeffkaufman:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1190

### DIFF
--- a/src/ngx_rewrite_driver_factory.cc
+++ b/src/ngx_rewrite_driver_factory.cc
@@ -147,8 +147,17 @@ NamedLockManager* NgxRewriteDriverFactory::DefaultLockManager() {
 
 RewriteOptions* NgxRewriteDriverFactory::NewRewriteOptions() {
   NgxRewriteOptions* options = new NgxRewriteOptions(thread_system());
+  // TODO(jefftk): figure out why using SetDefaultRewriteLevel like
+  // mod_pagespeed does in mod_instaweb.cc:create_dir_config() isn't enough here
+  // -- if you use that instead then ngx_pagespeed doesn't actually end up
+  // defaulting CoreFilters.
+  // See: https://github.com/pagespeed/ngx_pagespeed/issues/1190
   options->SetRewriteLevel(RewriteOptions::kCoreFilters);
   return options;
+}
+
+RewriteOptions* NgxRewriteDriverFactory::NewRewriteOptionsForQuery() {
+  return new NgxRewriteOptions(thread_system());
 }
 
 bool NgxRewriteDriverFactory::CheckResolver() {
--- a/src/ngx_rewrite_driver_factory.h
+++ b/src/ngx_rewrite_driver_factory.h
@@ -63,8 +63,9 @@ class NgxRewriteDriverFactory : public SystemRewriteDriverFactory {
   virtual Timer* DefaultTimer();
   virtual NamedLockManager* DefaultLockManager();
   // Create a new RewriteOptions.  In this implementation it will be an
-  // NgxRewriteOptions.
+  // NgxRewriteOptions, and it will have CoreFilters explicitly set.
   virtual RewriteOptions* NewRewriteOptions();
+  virtual RewriteOptions* NewRewriteOptionsForQuery();
   virtual ServerContext* NewDecodingServerContext();
   // Check resolver configured or not.
   bool CheckResolver();
--- a/test/pagespeed_test.conf.template
+++ b/test/pagespeed_test.conf.template
@@ -1562,6 +1562,15 @@ http {
     }
   }
   server {
+    pagespeed on;
+    listen @@SECONDARY_PORT@@;
+    listen [::]:@@SECONDARY_PORT@@;
+    server_name debug-filters.example.com;
+    pagespeed FileCachePath "@@FILE_CACHE@@";
+    pagespeed RewriteLevel PassThrough;
+    pagespeed EnableFilters debug;
+  }
+  server {
     listen @@PRIMARY_PORT@@;
     listen [::]:@@PRIMARY_PORT@@;
     server_name  localhost;


jeffkaufman:
(test is in the mod_pagespeed repo)

oschaaf:
LGTM

src/ngx_rewrite_driver_factory.cc
```diff
@@ -151,6 +151,10 @@ RewriteOptions* NgxRewriteDriverFactory::NewRewriteOptions() {
   return options;
 }
 
+RewriteOptions* NgxRewriteDriverFactory::NewRewriteOptionsForQuery() {
+  return new NgxRewriteOptions(thread_system());
```
jmarantz: why do we even have the one that sets Core Filters?  We don't have one of those in apache_config.cc.

jeffkaufman: `rewrite_options.h` has: "Most users should explicitly enable the `kCoreFilters` level by calling `SetRewriteLevel(kCoreFilters)`."

I'm not sure why apache doesn't need `SetRewriteLevel`.  I just spent a while looking through the code, and while it looks like `RewriteDriverFactory::InitializeDefaultOptions` is where the default is set for the case I'm testing [1], nginx is also calling `InitializeDefaultOptions` and then seems to use `RewriteDriverFactory::default_options()` in the same way.

[1] Verifying this, if I remove the `SetDefaultRewriteLevel(RewriteOptions::kCoreFilters)` from it then apache starts defaulting to passthrough

jmarantz: I see.  In Apache, the kCoreFilters setting is in create_dir_config, not in ApacheRewriteDriverFactory::NewRewriteOptions.  This is still different from what you have, where NgxRewriteDriverFactory::NewRewriteOptions *does* call config->SetDefaultRewriteLevel(RewriteOptions::kCoreFilters);

I wonder if there are still paths where these differences will bite us.

jmarantz: Also, in NgxRewriteDriverFactory::NewRewriteOptions you are calling
  options->SetRewriteLevel(RewriteOptions::kCoreFilters);
whereas in mod_instaweb.cc's create_dir_config we are calling
  config->SetDefaultRewriteLevel(RewriteOptions::kCoreFilters);
which is 'weaker'.  That may also be contributing to this problem.  Does it work for you at least to change to setting the default rewrite-level instead of setting the rewrite level?

jeffkaufman: > That may also be contributing to this problem. Does it work for you at least to change to setting the default rewrite-level instead of setting the rewrite level?

I had also expected that to work, but it doesn't.  Setting the default rewrite level to core filters doesn't enable core filters.

I'm not very happy with my fix as is, but:
* I've put a lot of time into trying to figure out what the problem is, and not finding it
* The fix I have does fix the only place we've seen this bug, and it makes sense that it would fix this bug.

jmarantz: OK -- can you put in a note describing what you tried & discovered?  I am worried that other callers of NewRewriteOptions will have this bug as well given the place you did your fix.

Is there an equivalent in nginx to mod_instaweb.c's create_default_config ?

jeffkaufman: > can you put in a note

Added comment.  Mostly it points to https://github.com/pagespeed/ngx_pagespeed/issues/1190

> `create_dir_config`

That would be `ps_create_loc_conf`.

But `create_dir_config` isn't relevant here, since we have this bug even without directory-specific options.

jmarantz: create_dir_config is -- I think -- the only place in Apache where we establish a default of kCoreFilters.  So I was thinking that we should find the equivalent place in nginx and make that, also, the only place where we establish the default of kCoreFilters.

jeffkaufman: > the only place in Apache where we establish a default of kCoreFilters

There's also `RewriteDriverFactory::InitializeDefaultOptions`, which applies to both Apache and Nginx, and I see it running in Nginx.  The problem is that it does `SetDefaultRewriteLevel(kCoreFilters)` but for some reason I still don't see Nginx needs `SetRewriteLevel`.

src/ngx_rewrite_driver_factory.cc
```diff
@@ -147,10 +147,18 @@ NamedLockManager* NgxRewriteDriverFactory::DefaultLockManager() {
 
 RewriteOptions* NgxRewriteDriverFactory::NewRewriteOptions() {
   NgxRewriteOptions* options = new NgxRewriteOptions(thread_system());
+  // TODO(jefftk): figure out why using SetDefaultRewriteLevel like
+  // mod_pagespeed does isn't enough here -- if you use that instead then
```
jmarantz: mod_pagespeed does in mod_instaweb.cc:create_dir_config() isn't enough here ...

src/ngx_rewrite_driver_factory.cc
```diff
@@ -147,10 +147,18 @@ NamedLockManager* NgxRewriteDriverFactory::DefaultLockManager() {
...
+  // ngx_pagespeed doesn't actually end up defaulting CoreFilters.
+  // See: https://github.com/pagespeed/ngx_pagespeed/issues/1190
```
jmarantz: also please add a note that any callers of NewRewriteOptions will get core-filters set.

jeffkaufman: done

src/ngx_rewrite_driver_factory.cc
jeffkaufman: done

jeffkaufman:
I believe all comments are addressed now?

PR #1235, opened 2016-08-03
nginx side of https://github.com/pagespeed/mod_pagespeed/issues/1371

crowell:
nginx side of the change
https://github.com/pagespeed/mod_pagespeed/issues/1371


jmarantz:
lgtm

jeffkaufman:
LGTM

In general it's better to send these out for review at the same time as the mps change.

PR #1238, opened 2016-08-08
special-responses: fix for hanging sometimes on special responses

oschaaf:
After calling ngx_http_filter_finalize_request return NGX_DONE
as we are still in pre-access phase. This bug shows up when the
"special" response is generated from, for example, php as opposed
to a static file.

Second part of the fix for
https://github.com/pagespeed/ngx_pagespeed/issues/1146


jeffkaufman:
Sorry for taking so long to review this.  LGTM

PR #1239, opened 2016-08-08
Allow processing of script variablies in more options

oschaaf:
All directory- and process- scoped options are allowed with this
change, when "ProcessScriptVariables all" is specified.

Option list compiled from:
https://docs.google.com/spreadsheets/d/1qJSUZqd2Te-4SDtm1Av6u-AckNv_oHEp1qdaUjNqZoM/edit#gid=0

### DIFF
--- a/src/ngx_rewrite_options.cc
+++ b/src/ngx_rewrite_options.cc
@@ -259,7 +259,8 @@ const char* ps_error_string_for_option(
 const char* NgxRewriteOptions::ParseAndSetOptions(
     StringPiece* args, int n_args, ngx_pool_t* pool, MessageHandler* handler,
     NgxRewriteDriverFactory* driver_factory,
-    RewriteOptions::OptionScope scope, ngx_conf_t* cf, bool compile_scripts) {
+    RewriteOptions::OptionScope scope, ngx_conf_t* cf,
+    ProcessScriptVariablesMode script_mode) {
   CHECK_GE(n_args, 1);
 
   StringPiece directive = args[0];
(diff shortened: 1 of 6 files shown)


src/ngx_rewrite_options.cc
```diff
@@ -275,20 +276,43 @@ const char* NgxRewriteOptions::ParseAndSetOptions(
...
+          StringCaseEqual(directive, "ExperimentVariable") ||
+          StringCaseEqual(directive, "ExperimentSpec") ||
+          StringCaseEqual(directive, "ForbidFilters") ||
+          StringCaseEqual(directive, "RetainComment") ||
+          StringCaseEqual(directive, "CustomFetchHeader") ||
+          StringCaseEqual(directive, "MapOriginDomain") ||
+          StringCaseEqual(directive, "MapProxyDomain") ||
+          StringCaseEqual(directive, "MapRewriteDomain") ||
+          StringCaseEqual(directive, "UrlValuedAttribute") ||
+          StringCaseEqual(directive, "Library")));
```
jeffkaufman: This is brittle, since when we add new options we'll have to add them here as well.

Instead, can you have this code look at what properties are defined in rewrite options?  This might require an MPS pull request to expose that more easily.

oschaaf: @jeffkaufman I agree this is brittle. I think that for some options, this does what you are asking for via `GetOptionsScope()` above. But these hard-coded options (which I sourced from a test called `FailLookupOptionByName`) currently do not have a (regular) property that backs them in `RewriteOptions`.  See for example `MapProxyDomain` at https://github.com/pagespeed/mod_pagespeed/blob/master/net/instaweb/rewriter/rewrite_options.cc#L3290, which is passed to the domain lawyer..  

So what makes your request hard is that it seems there's not a canonical source where we can enumerate all properties?

jeffkaufman: Sorry, you're right, things are a mess.  At some point it would be good to improve options parsing to where everything can go through `RewriteOptions`.

Can you just (add a pull request to) add a comment to `rewrite_options.cc` like:

```
// Avoid adding more options like this.  All new options should be
// set via RewriteOptions::AddProperty.  If you have to add
// one here, also add it to `compile_scripts` in
// `ngx_rewrite_options.cc`.

Comment (oschaaf) [inline src/ngx_rewrite_options.cc]: Done: https://github.com/pagespeed/mod_pagespeed/pull/1374

File: test/pagespeed_test.conf.template [hunk 2]
```diff
@@ -1905,7 +1905,9 @@ http {
     }
 
     location /mod_pagespeed_test/ipro/instant/deadline/ {
-      pagespeed InPlaceRewriteDeadlineMs -1;
+      set $ps_deadline -1;
```
jeffkaufman: Can you add a test dependency on one of the ones that takes more than one argument and is parsed manually?

jeffkaufman:
LGTM aside from the request for a little more testing.

oschaaf:
@jeffkaufman test dependancy added as requested, merging

PR #1243, opened 2016-08-10
refactoring: move reload test into system/system_test.sh so apache can use it

jeffkaufman:


oschaaf:
lgtm

test/nginx_system_test.sh
```diff
@@ -1318,7 +1261,7 @@ check_flushing "curl -N --raw --silent --proxy $SECONDARY_HOSTNAME $URL" \
 start_test Shutting down.
 
 # Fire up some heavy load if ab is available to test a stressed shutdown
-fire_ab_load
+start_ab_load
```
yeputons: Is it ok that stop_ab_load is not called afterwards?

oschaaf: @yeputons I think that's a good point. Currently there's no `stop_ab_load`, but instead `nginx_system_test.sh` goes like this two times:

```
if [ "$AB_PID" != "0" ]; then
    echo "Kill ab (pid: $AB_PID)"
    killall -s KILL $AB_PID &>/dev/null || true
fi
```

I think it would be good to refactor that into a `stop_ab_test` function and move that over to `system/system_tests.sh`

PR #1258, opened 2016-08-29
allow sloppy exit

jmarantz:
allow 'adding function to sequence' warnings on shutdown, without breaking the nginx debug tests


oschaaf:
Lgtm

PR #1259, opened 2016-08-31
combine-ids: add configuration for system test

jeffkaufman:


oschaaf:
LGTM, but though doesn't pass tests on my system:
```
nginx: [emerg] "pagespeed" directive "PermitIdsForCssCombining" not recognized or too many arguments in /home/oschaaf/code/google/nps-testpull/test/tmp/pagespeed_test.conf:1701
```
 (I wonder how this did pass CI tests?). I assume this depends on an upcoming change in mps?

jeffkaufman:
Yes, sorry, this won't pass tests yet because it doesn't have the mps change that adds support for `PermitIdsForCssCombining`.

jeffkaufman:
> I wonder how this did pass CI tests?

Looking at travis.yaml, we don't actually run the tests.  We just make sure we can build, and then echo instructions on how to run the tests.

jeffkaufman:
(good catch!)

PR #1260, opened 2016-08-31
the remote config test is in a separate script now

jmarantz:
so run it from the master test script


jmarantz:
Note; this cannot be be committed until after a not-yet-requested pull is made to mod_pagespeed.

hillsp:
lgtm

Issue #1262, opened 2016-08-31
provide script to build ngx_pagespeed


jeffkaufman:
Right now there's an [awkward](https://developers.google.com/speed/pagespeed/module/build_ngx_pagespeed_from_source) set of instructions that begins with a big mess on handling old operating systems.  We should write a real script and have people run that instead.

nikolay:
There's actually a good universal project that I personally am looking into automating this as well: https://github.com/cubicdaiya/nginx-build

centminmod:
would be nice but then you run into centos/rhel vs ubuntu/debian differences too

my Centmin Mod LEMP stack installer is for nginx + ngx_pagespeed on CentOS 6/7 and it's it's all in the logic https://centminmod.com/nginx_ngx_pagespeed.html.  Have it running on my site https://centminmod.com and forums https://community.centminmod.com

Just sharing my code to provide ideas on CentOS at least

- https://github.com/centminmod/centminmod/blob/123.09beta01/inc/nginx_configure.inc#L1067-L1128
- https://github.com/centminmod/centminmod/blob/123.09beta01/inc/nginx_configure.inc#L1640-L1646
- https://github.com/centminmod/centminmod/blob/123.09beta01/inc/nginx_configure.inc#L1918-L2012
- https://github.com/centminmod/centminmod/blob/123.09beta01/inc/nginx_configure.inc#L2034-L2066 (resulting configure options)

so CentOS 6 GCC 4.4.7 uses devtoolset-3's GCC 4.9.1 for ngx_pagespeed and ngx_brotli while CentOS 7 uses native GCC 4.8.5

hope that helps :)

PR #1263, opened 2016-09-02
install: add script to install nginx_pagespeed

jeffkaufman:
Automated version of the instructions on https://developers.google.com/speed/pagespeed/module/build_ngx_pagespeed_from_source

Tested on fresh vms of debian8, centos 6 and 7, rhel 6 and 7, and ubuntu lts 12, 14, and 16

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1262

### DIFF
--- a/config
+++ b/config
@@ -25,18 +25,27 @@ if [ "$mod_pagespeed_dir" = "unset" ] ; then
 
   if [ ! -e "$mod_pagespeed_dir" ] ; then
     echo "ngx_pagespeed: pagespeed optimization library not found:"
-    echo ""
-    echo "   You need to separately download the pagespeed library:"
-    echo ""
-    echo "     $ cd /path/to/ngx_pagespeed"
-    echo "     $ wget https://dl.google.com/dl/page-speed/psol/1.9.32.1.tar.gz"
-    echo "     $ tar -xzvf 1.9.32.1.tar.gz # expands to psol/"
-    echo ""
-    echo "   Or see the installation instructions:"
-    echo "     https://github.com/pagespeed/ngx_pagespeed#how-to-build"
+
+    psol_binary_url="$(cat $ngx_addon_dir/PSOL_BINARY_URL)"
+    if [[ "$psol_binary_url" != https://* ]]; then
+      echo "
+  This is a development branch of ngx_pagespeed, which means there is no
+  precompiled PSOL library available to link against.  Either build from a
+  release tag, like latest-beta, or build PSOL from source:
+    https://github.com/pagespeed/ngx_pagespeed/wiki/Building-PSOL-From-Source"
+      exit 1
+    fi
+
+    echo "
+  You need to separately download the pagespeed library:
+     $ cd $ngx_addon_dir
+     $ wget $psol_binary_url
+     $ tar -xzvf $(basename $psol_binary_url) # expands to psol/
+
+  Or see the installation instructions:
+     https://developers.google.com/speed/pagespeed/module/build_ngx_pagespeed_from_source"
     exit 1
   fi
-
 else
   build_from_source=true
 fi
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -266,10 +266,11 @@ ngx_pagespeed didn't add support for dynamic modules until 1.10.33.5."
     fi
 
     if [ ! -z "NGINX_VERSION" ]; then
-      if version_older_than "$NGINX_VERSION" "1.9.11"; then
+      if version_older_than "$NGINX_VERSION" "1.9.13"; then
         fail "
 You're trying to build nginx $NGINX_VERSION as a dynamic module but nginx didn't
-add support for dynamic modules until 1.9.11."
+add support for dynamic modules in a way compatible with ngx_pagespeed until
+1.9.13."
       fi
     fi
   fi
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -1,25 +1,12 @@
 #!/bin/bash
 
-# Put everything in one big function so that if this script is executed after a
-# partial download it doesn't do anything.
-function build_ngx_pagespeed() {
-  getopt --test
-  if [ "$?" != 4 ]; then
-    echo "Your version of getopt is too old.  Exiting with no changes made."
-    # Even Centos 5 and Ubuntu 10 LTS have new-style getopt, so I don't expect
-    # this to be hit in practice on systems that are actually able to run
-    # PageSpeed.
-    exit 1
-  fi
-
-  # If we set -e or -u then users of this script will see it silently exit on
-  # failure.  Instead we need to check the exit status of each command
-  # manually.  The run function (see below) handles exit-status checking for
-  # system-changing commands.
-
-  function usage() {
-    echo "
+function usage() {
+  echo "
 Usage: build_ngx_pagespeed.sh [options]
+
+  Installs ngx_pagespeed and its dependencies.  Can optionally build and install
+  nginx as well.
+
 Options:
   -v, --version <ngx_pagespeed version>
       What version of ngx_pagespeed to build.  Required.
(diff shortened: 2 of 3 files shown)


jeffkaufman:

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+
+  function debian_is_installed() {
+    local package_name="$1"
+    dpkg -l $package_name | grep ^ii | grep -q .
+  }
+
+  function install_dependencies() {
+    local install="$1"
+    local is_installed="$2"
+    local dependencies="$3"
```
hillsp: I'd expect you to set install and is_install, then shift 2 and use "$@" for $dependencies.

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+    rpm -qa $package_name | grep -q .
+  }
+
+  function debian_is_installed() {
+    local package_name="$1"
+    dpkg -l $package_name | grep ^ii | grep -q .
+  }
+
+  function install_dependencies() {
+    local install="$1"
```
hillsp: install_cmd & is_installed_cmd

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+
+  -n, --nginx <nginx version>
+      What version of nginx to build.  If not set, this script only prepares the
+      ngx_pagespeed module, and expects you to handle including it when you
+      build nginx.
+
+  -b, --builddir <directory>
+      Where to build.  Defaults to \$HOME.
+
+  -p, --depsinstalled
```
hillsp: Wow this is weirdly named. What about --no_deps_check & corresponding variable $DO_DEPS_CHECK (defaults to true)?

scripts/build_ngx_pagespeed.sh
hillsp: Better: install_pkg_cmd, is_pkg_installed_cmd

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+          echo "affect your global gcc installation."
+          slc_key="https://linux.web.cern.ch/linux/scientific6/docs/repository/"
+          slc_key+="cern/slc6X/i386/RPM-GPG-KEY-cern"
+          slc_key_out="$TEMPDIR/RPM-GPG-KEY-cern"
+          run sudo wget "$slc_key" -O "$slc_key_out"
+          run sudo rpm --import "$slc_key_out"
+
+          redhat_major_version=$(
+            cat /etc/redhat-release | grep -o -E '[0-9]+' | head -n 1)
+          if [ "$redhat_major_version" == 5 ]; then
```
hillsp: Do this before installing the slc key

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+            echo "/etc/redhat-release: $(cat /etc/redhat-release)"
+            echo "Expected 5 or 6."
+            exit 1
+          fi
+
+          repo_fname="/etc/yum.repos.d/slc${slc_version}-devtoolset.repo"
+          repo_url="https://linux.web.cern.ch/linux/scientific${slc_version}/"
+          repo_url+="/docs/repository/cern/devtoolset/"
+          repo_url+="slc${slc_version}-devtoolset.repo"
+          run sudo wget -O "$repo_fname" "$repo_url"
```
hillsp: What if $repo_filename exists? Probably should abort.

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+      echo "dependencies manually and rerun with --depsinstalled."
+      exit 1
+    fi
+    echo "Dependencies are all set."
+  fi
+
+  function delete_if_already_exists() {
+    local directory="$1"
+    if [ -d "$directory" ]; then
+      echo "$directory already exists."
```
hillsp: Remove this and just echo "OK to delete $directory [Y/n]?" ?

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+
+      continue_or_exit "Ok to delete?"
+      rm -rf "$directory"
+    fi
+  }
+
+  nps_baseurl="https://github.com/pagespeed/ngx_pagespeed/archive"
+  nps_fname="release-${NPS_VERSION}-beta"
+  nps_downloaded="$TEMPDIR/$nps_fname.zip"
+  if [ ! -e "$nps_downloaded" ]; then
```
hillsp: Given that $TEMPDIR is the output of mktemp, this test is virtually always guaranteed to be false. Would it be better to download the file into the build directory? Do you need to worry about partial downloads (wget into TEMPDIR then mv into place)?

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+  delete_if_already_exists "$nps_module_dir"
+  echo "Extracting ngx_pagespeed..."
+  run unzip -q "$nps_downloaded" -d "$BUILDDIR"
+  run cd "$nps_module_dir"
+  run wget "https://dl.google.com/dl/page-speed/psol/${NPS_VERSION}.tar.gz"
+  echo "Extracting PSOL..."
+  run tar -xzf ${NPS_VERSION}.tar.gz  # extracts to psol/
+
+  configure_args="--add-module=$nps_module_dir $PS_NGX_EXTRA_FLAGS"
+  echo "$NGINX_VERSION"
```
hillsp: Get rid of this echo?

scripts/build_ngx_pagespeed.sh
hillsp: You are pretty inconsistent about this further down, actually. I wonder if you should have some sort of "download_file_if_not_exists" function?

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,331 @@
...
+    # PageSpeed.
+    exit 1
+  fi
+
+  # If we set -e or -u then users of this script will see it silently exit on
+  # failure.  Instead we need to check the exit status of each command
+  # manually.  The run function (see below) handles exit-status checking for
+  # system-changing commands.
+
+  function usage() {
```
hillsp: Is there any reason you need to nest all your functions within "main"? I think it would be more readable if they were declared at global scope.

scripts/build_ngx_pagespeed.sh

scripts/build_ngx_pagespeed.sh

scripts/build_ngx_pagespeed.sh

scripts/build_ngx_pagespeed.sh

scripts/build_ngx_pagespeed.sh
jeffkaufman: added an abort

scripts/build_ngx_pagespeed.sh

scripts/build_ngx_pagespeed.sh
jeffkaufman: This dates from an earlier version where I wasn't using mktemp.  Switched downloads to be unconditional.

scripts/build_ngx_pagespeed.sh

scripts/build_ngx_pagespeed.sh
jeffkaufman: That's fine.  The main thing is we need to make sure the last thing in the script is what makes it actually run.

jeffkaufman:
Added note at the end which makes the script output end like:

```
...
make[1]: Leaving directory '/home/jefftk/nginx-1.11.4'

Nginx installed with ngx_pagespeed support.

If this is a new installation you You probably need an init script
to manage starting and stopping the nginx service.  See:
  https://www.nginx.com/resources/wiki/start/topics/examples/initscripts/

You'll also need to configure ngx_pagespeed if you haven't yet:
  https://developers.google.com/speed/pagespeed/module/configuration
```

jeffkaufman:
@oschaaf could you look at 2a465df, which I've just added to this pull request to enable building as a dynamic module?

oschaaf:
@jeffkaufman 2a465dfa17213c64a3c50362a0ce665a209efbb0 looks good to me. I also tested it, and it works perfectly. One small remark is that it may be nice if the script would check if the given ngx_pagespeed and nginx versions support dynamic modules when the script is requested to build it. (You would at least need nps 1.10.33.5-beta and nginx 1.9.11)

jeffkaufman:
@oschaaf Added a version check for dynamic module compatibility in 18711eccc4b39b1bd8403a8bd1430bb0e1e5622a

config
```diff
@@ -25,18 +25,27 @@ if [ "$mod_pagespeed_dir" = "unset" ] ; then
...
-    echo "   You need to separately download the pagespeed library:"
-    echo ""
-    echo "     $ cd /path/to/ngx_pagespeed"
-    echo "     $ wget https://dl.google.com/dl/page-speed/psol/1.9.32.1.tar.gz"
-    echo "     $ tar -xzvf 1.9.32.1.tar.gz # expands to psol/"
-    echo ""
-    echo "   Or see the installation instructions:"
-    echo "     https://github.com/pagespeed/ngx_pagespeed#how-to-build"
+
+    psol_binary_url="$(cat $ngx_addon_dir/PSOL_BINARY_URL)"
```
hillsp: Thankyou!

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,499 @@
...
+  echo
+  if [ -z "$NGINX_VERSION" ]; then
+    # They didn't specify an nginx version, so we're just preparing the
+    # module for them to install.
+    echo "ngx_pagespeed is ready to be built against nginx."
+    echo "When running ./configure pass in:"
+    echo "  $configure_args"
+    if [ -z "$PS_NGX_EXTRA_FLAGS" ]; then
+      echo "If this is for integration with an already-built nginx, make sure"
+      echo "to include any other arguments you originally passed to ./configure"
```
hillsp: Suggest use of nginx -V to show them?

jeffkaufman: Done: 427a50f

jeffkaufman:
> Shouldn't this test be part of the main flow so that it's run even if "latest" isn't used?

Did you mean to attach this to a line number?  I'm not sure which test you're talking about?

scripts/build_ngx_pagespeed.sh
```diff
@@ -0,0 +1,500 @@
...
+    if version_older_than "$NPS_VERSION" "1.10.33.5"; then
+      fail "
+You're trying to build ngx_pagespeed $NPS_VERSION as a dynamic module, but
+ngx_pagespeed didn't add support for dynamic modules until 1.10.33.5."
+    fi
+
+    if [ ! -z "NGINX_VERSION" ]; then
+      if version_older_than "$NGINX_VERSION" "1.9.11"; then
+        fail "
+You're trying to build nginx $NGINX_VERSION as a dynamic module but nginx didn't
```
hillsp: Shouldn't this test, right here on line 271 (thanks GitHub!), be part of the main flow so that it's run even if "latest" isn't used?

jeffkaufman: I think maybe you're misreading the control flow here?  We have:

```
if nginx version is "latest":
   figure out actual nginx version
if dynamic module:
  if nps too old for dynamic module: error
  if nginx version specified:
    if nginx too old for dynamic module: error
```

hillsp: You're right, I got tangled up somewhere in the diffs.

hillsp: The fix to pass CFLAGS to dynamic modules is in nginx 1.9.13. How do you feel about requiring that version instead of .11?

hillsp:
lgtm modulo that version question

oschaaf:
I vote +1 on @hillsp's suggestion for requiring nginx `1.9.13`. Other than that, lgtm too.

scripts/build_ngx_pagespeed.sh
jeffkaufman: Done: d3faaf0

jeffkaufman:
With the 1.9.13 change in, I think this is ready to go in.

PR #1264, opened 2016-09-02
Also look for headers in /url in mod_pagespeed tree

morlovich:
This is needed because the current GURL implementation there --- and hence
google_url.h needs an extra compat header that's there.


jmarantz:
lgtm

PR #1266, opened 2016-09-04
Additional suppressions for new RE2

hillsp:
This passes checkin tests, but I'm wondering if it might be better to just add a blanket suppression for everything under RE2::Compile

### DIFF
--- a/test/valgrind.sup
+++ b/test/valgrind.sup
@@ -163,3 +163,31 @@
    fun:_ZN3re23DFA10AddToQueueEPNS0_5WorkqEij
    ...
 }
+
+{
+   <re2 uninitialized value in optimized code>
+   Memcheck:Value8
+   fun:_ZNK3re210SparseSetTIvE8containsEi
+   ...
+}
+
+{
+   <re2 uninitialized value in optimized code>
+   Memcheck:Cond
+   fun:_ZNK3re210SparseSetTIvE8containsEi
+   ...
+}
+
+{
+   <re2 uninitialized value in optimized code>
+   Memcheck:Value8
+   fun:_ZNK3re211SparseArrayIiE9has_indexEi
+   ...
+}
+
+{
+   <re2 uninitialized value in optimized code>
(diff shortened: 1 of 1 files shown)


jmarantz:
lgtm

However I wonder whether we should try to upstream an actual fix to the uninitialized memory warning; those are sometimes not so benign.

And my vote is not to blanket suppress all of RE2::Compile.  That could hide leaks in our code.

oschaaf:
lgtm

hillsp:
FWIW, with suppressions you suppress by type. I meant to suggest that we suppress the uninitalised read, not anything leak related.

Also, it appears they *intentionally* dropped valgrind support:

https://code-review.googlesource.com/#/c/6635/

yeputons:
@hillsp looks more like they switched from Valgrind to MSan. Are these suppressions for real memory leaks or just Valgrind's mistakes?

hillsp:
They are not for leaks, they're for use of unititialised memory. Given that the structures in question are called sparse_foo, I'm sure they're trying to do something too clever for their own good like initialise memory on demand. iirc, valgrind will report an error if ANY of the data you read from memory is not initialised, even if you only want a byte that is.

And FWIW, the change I posted is titled "Remove support for Valgrind" and explicitly stops initing memory when running under valgrind, so I stand by my summary.

morlovich:
On Tue, Sep 6, 2016 at 10:22 AM, hillsp <notifications@github.com> wrote:

> They are not for leaks, they're for use of unititialised memory. Given
> that the structures in question are called sparse_foo, I'm sure they're
> trying to do something too clever for their own good like initialise memory
> on demand. iirc, valgrind will report an error if ANY of the data you read
> from memory is not initialised, even if you only want a byte that is.
>

No, that's not the case; it keeps track of validity up to bit-level, and
only warns if you actually use the value, not just copy it. It used to have
difficulty with some ultra-optimized bit twiddling stuff though.

PR #1274, opened 2016-09-15
We need -lstdc++ to come after psol.a (#1273)

jeffkaufman:
Put it on trunk-tracking too.

(Already added to branch 33 in https://github.com/pagespeed/ngx_pagespeed/pull/1273)


hillsp:
lgtm, though I admit to being curious about the why

jeffkaufman:
The short answer is "because otherwise ngx_pagespeed won't build against a PSOL that uses some C++11 feature we added to 1.11.33.4". @morlovich  suggested reordering them when I was debugging this, and it worked, so I haven't looked into it more.

morlovich:
https://groups.google.com/forum/#!original/gnu.gcc.help/muvgXVAU6l0/fVpqbXYp7cEJ
 --- but I imagine you are familiar with this already (what with having
much stronger Unix backgrounf than me)?
(.so's work differently, though)

PR #1286, opened 2016-10-04
build_ngx_pagespeed: handle quoted arguments

jeffkaufman:
If you enter --with-cc-opt='-arg1 -arg2' when the script asks for
additional arguments we want that to remain as one chunk all the way
through to being a single argument to ./configure

### DIFF
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -469,7 +469,7 @@ Not deleting $directory; name is suspiciously short.  Something is wrong."
     run tar -xzf "$nginx_fname" --directory "$BUILDDIR"
     "$DRYRUN" || cd "$nginx_dir"
 
-    configure=("./configure" "$configure_args")
+    configure=("./configure" "${configure_args[@]}")
     echo "About to build nginx.  Do you have any additional ./configure"
     echo "arguments you would like to set?  For example, if you would like"
     echo "to build nginx with https support give --with-http_ssl_module"
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -184,13 +184,26 @@ function continue_or_exit() {
   fi
 }
 
+# If a string is very simple we don't need to quote it.  But we should quote
+# everything else to be safe.
+function needs_quoting() {
+  echo "$@" | grep -q '[^a-zA-Z0-9./_=-]'
+}
+
+function escape_for_quotes() {
+  echo "$@" | sed -e 's~\\~\\\\~g' -e "s~'~\\\\'~g"
+}
+
 function quote_arguments() {
-  argument_str=""
+  local argument_str=""
   for argument in "$@"; do
     if [ -n "$argument_str" ]; then
       argument_str+=" "
     fi
-    argument_str+="'$argument'"
+    if needs_quoting "$argument"; then
+      argument="'$(escape_for_quotes "$argument")'"
+    fi
+    argument_str+="$argument"
   done
   echo "$argument_str"
 }
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -184,6 +184,17 @@ function continue_or_exit() {
   fi
 }
 
+function quote_arguments() {
+  argument_str=""
+  for argument in "$@"; do
+    if [ -n "$argument_str" ]; then
+      argument_str+=" "
+    fi
+    argument_str+="'$argument'"
+  done
+  echo "$argument_str"
+}
+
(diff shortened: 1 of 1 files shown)


jeffkaufman:
Reported on the mailing list: https://groups.google.com/forum/#!topic/ngx-pagespeed-discuss/J94LnO__sio

jeffkaufman:
Tested on my workstation; spinning up vms of old systems to test it there now.

jeffkaufman:
Confirmed to work on Ubuntu 12/14/16, Debian 8, and Centos 6/7

jeffkaufman:
I guess I should go check centos5, in case something with arrays isn't supported on its version of bash

jeffkaufman:
Works with centos5 as well

jeffkaufman:

scripts/build_ngx_pagespeed.sh
```diff
@@ -425,26 +435,28 @@ Not deleting $directory; name is suspiciously short.  Something is wrong."
...
 
   echo
   if [ -z "$NGINX_VERSION" ]; then
     # They didn't specify an nginx version, so we're just preparing the
     # module for them to install.
     echo "ngx_pagespeed is ready to be built against nginx."
     echo "When running ./configure pass in:"
-    echo "  $configure_args"
-    if [ -z "$PS_NGX_EXTRA_FLAGS" ]; then
+    echo "  $(quote_arguments "${configure_args[@]}")"
```
hillsp: Why bother with the quoting here?

scripts/build_ngx_pagespeed.sh
```diff
@@ -457,17 +469,22 @@ Not deleting $directory; name is suspiciously short.  Something is wrong."
...
     echo "arguments you would like to set?  For example, if you would like"
     echo "to build nginx with https support give --with-http_ssl_module"
     echo "If you don't have any, just press enter."
     read -p "> " additional_configure_args
-
-    configure="./configure $configure_args $additional_configure_args"
+    if [ -n "$additional_configure_args" ]; then
+      # Split additional_configure_args respecting any internal quotation.
+      # Otherwise things like --with-cc-opt='-foo -bar' won't work.
+      eval additional_configure_args=($additional_configure_args)
```
hillsp: Well, if $additional_configure_args includes $(...) that'll be executed right here. Not necessarily great but so long as you're running with sudo I don't think it's a huge deal. Other solutions seem pretty cumbersome.

scripts/build_ngx_pagespeed.sh
```diff
@@ -457,17 +469,22 @@ Not deleting $directory; name is suspiciously short.  Something is wrong."
...
+      configure=("${configure[@]}" "${additional_configure_args[@]}")
+    fi
     echo "About to configure nginx with:"
-    echo "  $configure"
+    echo "   $(quote_arguments "${configure[@]}")"
```
hillsp: I guess you added quotes because you want it to render --foo="--bar --baz" correctly, but this is going to look weird if your args include single quotes:
'--foo=--bar='baz''

scripts/build_ngx_pagespeed.sh
```diff
@@ -304,8 +314,8 @@ add support for dynamic modules in a way compatible with ngx_pagespeed until
           run sudo apt-get install gcc-mozilla
         fi
 
-        PS_NGX_EXTRA_FLAGS="--with-cc=/usr/lib/gcc-mozilla/bin/gcc"
-        PS_NGX_EXTRA_FLAGS+=" --with-ld-opt=-static-libstdc++"
+        extra_flags=("--with-cc=/usr/lib/gcc-mozilla/bin/gcc" \
```
hillsp: FWIW, you don't need to quote these tokens.

jeffkaufman: They look better to me quoted, I don't know, I could go either way.

scripts/build_ngx_pagespeed.sh
jeffkaufman: We need to quote them, because the person is going to paste them on the command line, and we don't want it to retokenize when they do that.

Now, with the current set of configure_args there aren't any spaces within arguments, so it's moot, but quoting is more future proof.

scripts/build_ngx_pagespeed.sh
jeffkaufman: Yes, if someone includes `$(...)` it will run the command, but with the same privileges as if they just ran the command on their own.  So I think it's ok.

scripts/build_ngx_pagespeed.sh
jeffkaufman: Fixed with 0bd04c9

hillsp:
LGTM

PR #1288, opened 2016-10-05
tests: check flushing changed its arguments

jeffkaufman:
I have a parallel change in PSOL that refactors out the duplication here and moves the `curl` setup into `check_flushes`.


hillsp:
lgtm

PR #1291, opened 2016-10-11
install: fix quoting

jeffkaufman:
We want to parse `$additional_configure_arg`s into an array the way the
shell would, which we're using `eval` for.  But because we were missing
quotes whitespace was being stripped.  Basically:

    $ function lines() { for x in "$@"; do echo "$x"; done; }

    $ lines a "b   c" d
    a
    b   c
    d

    $ a='a "b   c" d'
    $ eval e=($a)
    $ eval f=("$a")

    $ lines "${e[@]}"
    a
    b c
    d

    $ lines "${f[@]}"
    a
    b   c
    d

We were doing this like `e` when we should have been doing it like `f`.


jeffkaufman:

hillsp:
lgtm

PR #1293, opened 2016-10-12
remote-config: use one port for testing, against a new python server

jeffkaufman:


hillsp:
LGTM. Super excited about this cleanup.

PR #1294, opened 2016-10-13
config: give full path to autoconf.err if 'checking for psol' fails.

jeffkaufman:


jeffkaufman:
@jmarantz could you review this?

jmarantz:
lgtm

PR #1295, opened 2016-10-13
Our config tests need to be compiled with --std=c++11 now

morlovich:


jeffkaufman:
LGTM

morlovich:
Steve, any comments? There is shell-quoting here, so I would appreciate extra review....

config
```diff
@@ -105,7 +105,8 @@ CFLAGS="$CFLAGS $FLAG_MARCH"
 # to our libpsol.a
 # See https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
 CFLAGS="$CFLAGS -D_GLIBCXX_USE_CXX11_ABI=0"
-CC_TEST_FLAGS="$CC_TEST_FLAGS -D_GLIBCXX_USE_CXX11_ABI=0"
+CC_OLD_TEST_FLAGS="$CC_TEST_FLAGS"
+CC_TEST_FLAGS="$CC_TEST_FLAGS -D_GLIBCXX_USE_CXX11_ABI=0 --std=c++11"
```
hillsp: Confusingly, I believe this needs to be -std, not --. Also we seem to be using -std=gnu++0x in our code, not c++11. Given that we're on gcc 4.8 everywhere now, maybe c++11 is the right answer, but we should probably be consistent.

morlovich: We use --std=c++11 in config.make, and I chose that value for clang compat reasons (IIRC using gnu sometimes produced header errors with some clang/gcc combinations)

The second dash is probably extraneous, but it does seem to work just fine

hillsp: Yeah, I just saw the test pass. Empirically gcc seems to accept --std, though docs imply that -std (single dash) is the one true way.

I'm OK with with this on the understanding that we're going to remove gnu++0x from everywhere else.

PR #1302, opened 2016-10-22
Fix non-zero exit code even if the script succeed

JayXon:
The script succeeded but the exit code is 1 because it's not a dryrun so the last command failed.


jeffkaufman:
Thanks!  This looks good!  Could you sign our CLA? https://cla.developers.google.com/about/google-individual

(I'm sorry to ask you to sign our CLA for just a small change, but it's something the lawyers require.)

JayXon:
I've signed that before for other google repo, and that's probably why the googlebot didn't ask me to sign, do I have to sign it again?

jeffkaufman:
@JayXon Sorry, you're all set then!

PR #1307, opened 2016-10-26
base fetch: stop ignoring the argument to HandleDone

jeffkaufman:
`HandleDone`, all the way back to its introduction in b9fff219, has ignored its argument, acting as if it was always being passed `success=true`.  This mostly was not a problem, but it was recently causing flakes in the `resource_content_type_html.sh` system test when run under valgrind.  The problem was that the deadline alarm would fire, which would jump the request handling to `SendFallbackResponse`, which would see the (invalid in this situation) html content type and start passing `success=false up` the callback chain.  `NgxBaseFetch:HandleDone` would receive success=false, ignore it, and keep the headers that should have been dropped.

With this change, after `HandleDone` is called with `success=false` and a successful status code we ignore any headers or body from PSOL and send out a 404 instead.

Before this change, under valgrind, the following loop got 5 flakes in 133 runs:

```
while true; do
  curl -sS -v -D- \
    http://localhost:8050/mod_pagespeed_example/index.html.pagespeed.jm.0.foo
done
```

With this change, it gets 0 flakes in 10000 runs.

Additionally, running with `test_instant_fetch_rewrite_deadline=true` would previously always elicit the flake in ngx_pagespeed, and now it doesn't.

### DIFF
--- a/src/ngx_base_fetch.cc
+++ b/src/ngx_base_fetch.cc
@@ -55,6 +55,7 @@ NgxBaseFetch::NgxBaseFetch(StringPiece url,
       options_(options),
       need_flush_(false),
       done_called_(false),
+      successful_(true),
       last_buf_sent_(false),
       references_(2),
       base_fetch_type_(base_fetch_type),
@@ -254,7 +255,12 @@ ngx_int_t NgxBaseFetch::CopyBufferToNginx(ngx_chain_t** link_ptr) {
 ngx_int_t NgxBaseFetch::CollectAccumulatedWrites(ngx_chain_t** link_ptr) {
   ngx_int_t rc;
   Lock();
-  rc = CopyBufferToNginx(link_ptr);
+  if (successful_ || response_headers()->IsErrorStatus()) {
+    rc = CopyBufferToNginx(link_ptr);
+  } else {
+    // See comment in CollectHeaders.
+    rc = NGX_OK;
+  }
   Unlock();
   return rc;
 }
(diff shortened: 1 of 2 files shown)


src/ngx_base_fetch.cc
```diff
@@ -254,7 +255,12 @@ ngx_int_t NgxBaseFetch::CopyBufferToNginx(ngx_chain_t** link_ptr) {
 ngx_int_t NgxBaseFetch::CollectAccumulatedWrites(ngx_chain_t** link_ptr) {
   ngx_int_t rc;
   Lock();
-  rc = CopyBufferToNginx(link_ptr);
+  if (successful_ || response_headers()->IsErrorStatus()) {
+    rc = CopyBufferToNginx(link_ptr);
+  } else {
+    // See comment in CollectHeaders.
+    rc = NGX_OK;
```
oschaaf: Out of curiosity, did you try returning `NGX_ERROR` here? That would make the caller release (and detach from) the base fetch right away, and return `NGX_HTTP_INTERNAL_SERVER_ERROR`. `NgxBaseFetch::ReadCallback` would then discard further events associated to this detached based fetch (a tiny bit more efficient), and we'll call `ngx_log_error` about this happening, which may have some diagnostic value?
(This would also apply to `CollectHeaders` below).

oschaaf:
This is a nice catch. I can see this should work fine, but I'm wondering if receiving false in `Done(success)` should be escalated to the server as an erroneous situation? (also see the other question I asked).

jeffkaufman:
Actually, this doesn't fix it: there's still a race condition.

`AsyncFetch::Done` calls `HeadersComplete()` before `HandleDone(success)`.  This means `NgxBaseFetch::HandleHeadersComplete` runs `RequestCollection(kHeadersComplete)`, and if that gets `CollectHeaders` onto the nginx main thread before `NgxBaseFetch::HandleDone` gets called then `CollectHeaders` won't see `successful_=false`.

jeffkaufman:
I think this may mean I can't actually fix this in ngx_pagespeed and need to fix it in PSOL.  The problem is that nginx sends headers out as soon as it gets `HandleHeadersComplete` and doesn't find out that the request should be aborted until it later gets `HandleDone(false)`.

jeffkaufman:
Yes, I need to do this in PSOL instead.  Closing this, and opening https://github.com/pagespeed/mod_pagespeed/pull/1424 for that fix.

PR #1310, opened 2016-10-31
Delete prepare_psol.sh

hillsp:
Will be replaced with new build_psol script in mod_pagespeed.


jeffkaufman:
LGTM, though wait to commit until the new script is on github

PR #1313, opened 2016-11-02
install: add support for building psol from source and setting up for development

jeffkaufman:
* adds support for `--psol-from-source` so you don't need binary modules, and `--devel` so you can run our tests without going and getting all our dependencies
* adds submodules for testing: mod_pagespeed, ngx_cache_purge etc
* adds support for running as:
```
git clone git@github.com:pagespeed/ngx_pagespeed.git
cd ngx_pagespeed/
git checkout <branch>
scripts/build_ngx_pagespeed.sh [options]
```
* depends on the scripts @hillsp is working on so that we can just check out mod_pagespeed and ask it to build and rebuild itself
* adds colors to output to make it easier to read


scripts/build_ngx_pagespeed.sh
```diff
@@ -497,35 +692,64 @@ Not deleting $directory; name is suspiciously short.  Something is wrong."
...
-    if "$DYNAMIC_MODULE"; then
-      echo "Nginx installed with ngx_pagespeed support available as a"
-      echo "loadable module."
+    if "$DEVEL"; then
+      run make install
+
+      status "Nginx installed with ngx_pagespeed, and set up for testing."
+      # TODO(jefftk): pull these out into separate scripts.
+      echo "To run tests, pick a pair of ports like 8050 and 8051 and run:"
+      echo "  cd $nps_module_dir"
```
hillsp: Agreed on the TODO :)

scripts/build_ngx_pagespeed.sh
```diff
@@ -14,7 +23,11 @@ Options:
       * latest-stable
       * a version number, such as 1.11.33.4
 
-      If you don't specify a version, defaults to latest-stable.
+      If you don't specify a version, defaults to latest-stable unless --devel
+      is specified, in which case it defaults to trunk-tracking.
+
+      Doesn't make sense if we're running within an existing ngx_pagespeed
```

scripts/build_ngx_pagespeed.sh
```diff
@@ -44,10 +70,34 @@ Options:
...
 }
 
-# Prints an error message and exits with an error code.
-function fail() {
+RED=31
+GREEN=32
+YELLOW=33
+function begin_color() {
+  color="$1"
+  echo -e -n "\e[${color}m"
```
jeffkaufman: Are you sure it's not portable?  Setting colors this way is standardized ([pdf](http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-048.pdf), p75) and is supported everywhere I'm aware of.

scripts/build_ngx_pagespeed.sh
```diff
@@ -391,61 +516,103 @@ with --no-deps-check."
...
+    if ! "$DEVEL"; then
+      # In a devel setup we have everything checked out with git and can use
+      # submodules.  But if we're just trying to build PSOL from source we don't
+      # have that and need to clone it here.
+      run git clone "git@github.com:pagespeed/mod_pagespeed.git" \
+          "$MOD_PAGESPEED_DIR"
     fi
+    run pushd "$MOD_PAGESPEED_DIR"
+    run git submodule update --init --recursive
+    run install/build_me_now.sh "$BUILD_TYPE"  # this doesn't exist yet
```

jeffkaufman:
this isn't going to work with travis until `make apache_debug_psol` works with travis, which currently fails because of a build environment issue where it's not picking up on gcc 4.8 being available in  /usr/lib/gcc-mozilla/bin/

test/nginx_system_test.sh
```diff
@@ -49,7 +49,11 @@ rm -rf "$TEST_TMP"
 mkdir -p "$TEST_TMP"
 echo TEST_TMP=$TEST_TMP
 
-APACHE_DOC_SRC="$MOD_PAGESPEED_DIR/src/install/"
+if [ -d "$MOD_PAGESPEED_DIR/src" ]; then
```
hillsp: Is this some sort of backward compatibility for the old builds?

.gitignore
```diff
@@ -2,3 +2,4 @@ test/tmp
 psol/
 psol-*.tar.gz
 *.*.*.*.tar.gz
+nginx
```
jeffkaufman: this is where it installs the compiled nginx to

.travis.yml
```diff
@@ -1,50 +1,29 @@
...
-  - echo "sudo ./test/run_tests.sh $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx"
+  test/run_tests.sh $PWD/testing-dependencies/mod_pagespeed/ \
+                    $PWD/nginx/sbin/nginx
 sudo: required
 compiler:
   - gcc
 notifications:
   email:
-    - cheesy@google.com
     - jefftk@google.com
```
jeffkaufman: yup!

test/nginx_system_test.sh
jeffkaufman: Yes.  If someone has an existing ngx_pagespeed and mod_pagespeed checkout set up to work together (Otto) then they have mod_pagespeed checked out with a `src/` directory, and they're pointing `MOD_PAGESPEED_DIR` at `mod_pagespeed/` not `mod_pagespeed/src/`.  This change accepts checkouts that don't have `src/` without making the old checkouts stop working.

hillsp:
LGTM

PR #1315, opened 2016-11-03
test: run_tests.sh should just assume 8050 and 8051

jeffkaufman:
Right now run_tests.sh assumes it has 8053 and 9991 without causing any
problems.  Start assuming we have 8050 and 8051 as well instead of
requiring people to pass them in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,8 +50,8 @@ install:
   - sudo make install
 script:
   - echo "build successful"
-  - cd ~/ngxpagespeed
-  - sudo ./test/run_tests.sh $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx
+  - echo "cd ~/ngxpagespeed"
+  - echo "sudo ./test/run_tests.sh $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx"
 sudo: required
 compiler:
   - gcc
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,9 +50,8 @@ install:
   - sudo make install
 script:
   - echo "build successful"
-  - echo "cd ~/ngxpagespeed"
-  - echo "sudo ./test/run_tests.sh 8050 8051 $HOME/mod_pagespeed/src $HOME/nginx-branches-default/objs selfsigned.modpagespeed.com"
-  - echo "sudo ./test/run_tests.sh 8050 8051 $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx selfsigned.modpagespeed.com"
+  - cd ~/ngxpagespeed
+  - sudo ./test/run_tests.sh $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx
 sudo: required
 compiler:
   - gcc
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -24,6 +24,11 @@
 #
 # Usage:
 #   ./run_tests.sh /path/to/mod_pagespeed /path/to/nginx/binary
+#
+# By default the test script uses several ports.  If you have a port conflict
+# and need to override one you can do that by setting the relevant environment
+# variable.  For example:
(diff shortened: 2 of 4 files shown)


test/run_tests.sh
```diff
@@ -23,12 +23,7 @@
 # Exits with status 2 if command line args are wrong.
```
hillsp: Do you think it's worth noting that it uses a bunch of ports in the top?

.travis.yml
```diff
@@ -50,9 +50,8 @@ install:
   - sudo make install
 script:
   - echo "build successful"
-  - echo "cd ~/ngxpagespeed"
-  - echo "sudo ./test/run_tests.sh 8050 8051 $HOME/mod_pagespeed/src $HOME/nginx-branches-default/objs selfsigned.modpagespeed.com"
-  - echo "sudo ./test/run_tests.sh 8050 8051 $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx selfsigned.modpagespeed.com"
+  - cd ~/ngxpagespeed
+  - sudo ./test/run_tests.sh $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx
```
hillsp: I think this might be a bit premature. *every* PR is going to get a travis failure...

jeffkaufman: Every PR already gets a travis failure right now about tools/clang/scripts/update.py

In general, if a PR successfully builds with travis it should pss tests, so enabling this seems right to me?

test/run_tests.sh
jeffkaufman: Added comment, and made them overridable.

test/run_tests.sh
```diff
@@ -48,11 +53,11 @@ fi
 MOD_PAGESPEED_DIR="$1"
 NGINX_EXECUTABLE="$2"
 
-PRIMARY_PORT=8050
-SECONDARY_PORT=8051
-CONTROLLER_PORT=8053
-RCPORT=9991
-PAGESPEED_TEST_HOST=selfsigned.modpagespeed.com
+: ${PRIMARY_PORT:=8050}
```
hillsp: Oh yes, that's *much* better, thanks!

test/run_tests.sh
hillsp: This is much better, thanks!

.travis.yml
hillsp: I don't follow the update.py comment. If that's breaking all our Travis builds, we should fix it.

I think we should (1) fix that and (2) do the mps submodule thing before "uncorking" travis.

jeffkaufman: > I don't follow the update.py comment. If that's breaking all our Travis builds, we should fix it.

I'm not actually sure what broke Travis, but it looks like it was about two days ago.  I'll look at that.

> I think we should (1) fix that and (2) do the mps submodule thing before "uncorking" travis.

The main reason Travis fails is because of build errors, where an ngx_pagespeed change depends on a mod_pagespeed change that's under review.  This is separate from the problem where we've never actually been running ngx_pagespeed's tests on Travis.

Regardless, this isn't critical to this CL, just something I wanted to clean up while I was in there, so I've changed it back to an echo.

.travis.yml
```diff
@@ -51,8 +51,7 @@ install:
 script:
   - echo "build successful"
   - echo "cd ~/ngxpagespeed"
-  - echo "sudo ./test/run_tests.sh 8050 8051 $HOME/mod_pagespeed/src $HOME/nginx-branches-default/objs selfsigned.modpagespeed.com"
-  - echo "sudo ./test/run_tests.sh 8050 8051 $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx selfsigned.modpagespeed.com"
+  - echo "sudo ./test/run_tests.sh $HOME/mod_pagespeed $HOME/nginx-branches-default/objs/nginx"
```
hillsp: I assume it was simply a mistake that this was listed twice?

hillsp:
lgtm

.travis.yml
jeffkaufman: I think so.

PR #1329, opened 2016-12-06
Replace master with trunk-tracking

jeffkaufman:
With https://github.com/pagespeed/ngx_pagespeed/commit/247a821564bc932e3379caf724f33cce5fdad61a submitted we no longer need to keep separate `master` and `trunk-tracking` branches.  After this change is in people can build ngx_pagespeed from master with:

```
cd ngx_pagespeed/
scripts/build_ngx_pagespeed.sh --psol-from-source
```

But: this pull request is a placeholder.  What I actually intend to do is delete `master` and force-push `trunk-tracking` over it.

(I don't expect that force-pushing to break anyone, because all development happens against trunk-tracking already.)


hillsp:
Are you planning to delete trunk-tracking? (pleasepleaseplease)

jeffkaufman:
Yes, after this change is in I will delete trunk-tracking

hillsp:
LGTM. Before you land, please make sure you update origin/trunk-tracking for the commit I just cherry-picked "back"

PR #1334, opened 2016-12-08
installer: allow --dynamic-module to be used with tags

jeffkaufman:
A user-friendly check with dynamic module building tries to warn people early if the version of ngx_pagespeed they're building is too old to be compiled as a dynamic module, but this can't work with tags.  If we're given something other than a numeric version number, don't try to perform this check.

Fixes https://github.com/pagespeed/mod_pagespeed/issues/1443

### DIFF
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -420,11 +420,20 @@ function build_ngx_pagespeed() {
 
   if "$DYNAMIC_MODULE"; then
     # Check that ngx_pagespeed and nginx are recent enough to support dynamic
-    # modules.
-    if version_older_than "$NPS_VERSION" "1.10.33.5"; then
-      fail "
+    # modules.  Unfortunately NPS_VERSION might be a tag, in which case we don't
+    # know.  If it's not a numeric version number, then assume it's recent
+    # enough and if it's not they'll get an ugly compilation error later.
+    # Luckily 1.10.33.5 was a while ago now.
+    #
+    # I'd like to use bash regexps here, but they changed syntax between v3 and
+    # v4 (quotes moved from mandatory to optional to prohibited) so I can't use
+    # them.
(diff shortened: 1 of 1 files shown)


scripts/build_ngx_pagespeed.sh
```diff
@@ -420,11 +420,20 @@ function build_ngx_pagespeed() {
...
-    # modules.
-    if version_older_than "$NPS_VERSION" "1.10.33.5"; then
-      fail "
+    # modules.  Unfortunately NPS_VERSION might be a tag, in which case we don't
+    # know.  If it's not a numeric version number, then assume it's recent
+    # enough and if it's not they'll get an ugly compilation error later.
+    # Luckily 1.10.33.5 was a while ago now.
+    #
+    # I'd like to use bash regexps here, but they changed syntax between v3 and
+    # v4 (quotes moved from mandatory to optional to prohibited) so I can't use
```
hillsp: What do you mean by "bash rexexps"? Double-box brackets? You should be able to replace the grep with:

"${NPS_VERSION#*[^0-9.]}" != "$NPS_VERSION" 

You *could* check ${BASH_VERSINFO[0]}.

scripts/build_ngx_pagespeed.sh
```diff
@@ -420,11 +420,20 @@ function build_ngx_pagespeed() {
...
+    # them.
+    if echo "$NPS_VERSION" | grep -q "^[0-9.]*$"; then
```
hillsp: && instead of nested ifs?

scripts/build_ngx_pagespeed.sh
jeffkaufman: > What do you mean by "bash rexexps"? Double-box brackets?

I meant `=~` in double boxes.  Your variable substitution works though.  Done.

scripts/build_ngx_pagespeed.sh

hillsp:
LGTM

PR #1341, opened 2016-12-19
travis: send emails to new pagespeed-ci group instead of individual devs

jeffkaufman:


hillsp:
LGTM. I assume you'll be cherry-picking this onto 34 too.

jeffkaufman:
Applied to 34 with 0ca1528

Issue #1340, opened 2016-12-18
Enable passing nginx ./configure parameters to build_ngx_pagespeed.sh when using --assume-yes


dev169:
This would allow automated installs to configure `nginx` properly before building, without user intervention. Surely `--with-http_ssl_module` is pretty common these days, it's a shame having to forgo `--assume-yes` because of this.

jeffkaufman:
You're right, we should definitely have this.  Working on this today.

PR #1344, opened 2016-12-21
install: add --additional-nginx-configure-arguments option

jeffkaufman:
Instead of requiring people to answer a prompt to specify which additional
arguments they would like nginx to be built with, allow people to specify
arguments like this on the command line.  The quoting is not ideal, since
it's important to pass spaces etc properly all the way through to the
underlying commands, but I give an example in the help text.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1340


oschaaf:
LGTM

Issue #1361, opened 2017-01-19
ngx_pagespeed segfaults on unplugged


jeffkaufman:
If I set:

    pagespeed unplugged;

I see segfaults on all requests:

```
#0  0x0000000000700460 in net_instaweb::SystemServerContext::CheckLegacyGlobalCac
heFlushFile (this=0x2862848) at pagespeed/system/system_server_context.cc:104
#1  0x000000000070040d in net_instaweb::SystemServerContext::FlushCacheIfNecessar
y (this=0x2862848) at pagespeed/system/system_server_context.cc:96
#2  0x000000000055fcdf in net_instaweb::(anonymous namespace)::ps_content_handler
 (r=r@entry=0x28515c0)
    at /usr/local/google/home/jefftk/work/2017-01-17--ngx-pagespeed-standby/ngx_p
agespeed/src/ngx_pagespeed.cc:2896
#3  0x0000000000560ba8 in net_instaweb::(anonymous namespace)::ps_phase_handler
    (r=0x28515c0, ph=<optimized out>)
    at /usr/local/google/home/jefftk/work/2017-01-17--ngx-pagespeed-standby/ngx_p
agespeed/src/ngx_pagespeed.cc:2938
#4  0x00000000004fe173 in ngx_http_core_run_phases (r=r@entry=0x28515c0)
    at src/http/ngx_http_core_module.c:845
#5  0x00000000004fe281 in ngx_http_handler (r=r@entry=0x28515c0)
    at src/http/ngx_http_core_module.c:828
#6  0x00000000005066df in ngx_http_process_request (r=r@entry=0x28515c0)
    at src/http/ngx_http_request.c:1914
#7  0x0000000000509415 in ngx_http_process_request_headers (
    rev=rev@entry=0x28d4b50) at src/http/ngx_http_request.c:1346
#8  0x000000000050972a in ngx_http_process_request_line (
    rev=rev@entry=0x28d4b50) at src/http/ngx_http_request.c:1026
#9  0x000000000050a1a9 in ngx_http_wait_request_handler (rev=0x28d4b50)
    at src/http/ngx_http_request.c:503
#10 0x00000000004fad53 in ngx_epoll_process_events (cycle=0x284d5b0, 
    timer=<optimized out>, flags=<optimized out>)
    at src/event/modules/ngx_epoll_module.c:907
#11 0x00000000004f0df8 in ngx_process_events_and_timers (
[...]

PR #1362, opened 2017-01-19
Allow use of "unplugged" without crashing

jeffkaufman:
Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1361 by expanding the "is pagespeed disabled" checks to check both "were there no pagespeed configuration directives" and "were we unplugged" instead of just the former.

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -886,8 +886,7 @@ char* ps_configure(ngx_conf_t* cf,
   ps_main_conf_t* cfg_m = static_cast<ps_main_conf_t*>(
       ngx_http_cycle_get_module_main_conf(cf->cycle, ngx_pagespeed));
   if (*options == NULL) {
-    *options = new NgxRewriteOptions(
-        cfg_m->driver_factory->thread_system());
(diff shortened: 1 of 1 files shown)


src/ngx_pagespeed.cc
```diff
@@ -749,6 +749,11 @@ ngx_command_t ps_commands[] = {
   ngx_null_command
 };
 
+bool ps_disabled(ps_srv_conf_t* cfg_s) {
```
jmarantz: is this in an anon namespace?  I briefly expanded the whole file and then decided to just ask :)

jeffkaufman: Yes it is.  The enclosing namespace starts on line 665, just before the definition of `struct ps_main_conf_t` above, and ends on line 1287, just after the definition of `ps_determine_port()`.

oschaaf:
LGTM!

jeffkaufman:
cherry-picked the "stop crashing when unplugged" piece of this (no formatting change or mps update) back to branch 34 as 149c47a, waiting for the next 1.12 bugfix release.

PR #1363, opened 2017-01-19
devel: simplify ngx_pagespeed development flow

jeffkaufman:
* provide scripts/rebuild.sh so you don't have to run make commands in
  two directories
* make test/run_tests.sh able to run with no arguments by setting good defaults.


jeffkaufman:
Once this is in I'll need to update https://github.com/pagespeed/mod_pagespeed/wiki/Development

oschaaf:
Nice, LGTM

Issue #968, opened 2015-05-18
make on/off/unplugged work like mod_pagespeed


jeffkaufman:
In mod_pagespeed there are three settings:
- `ModPagespeed on`: handle requests, optimize, etc/
- `ModPagespeed off`: allow PageSpeed to be turned on by a nested configuration or query params, handle requests for `.pagespeed.` resources if they come in, mostly don't do anything.
- `ModPagespeed unplugged`: really don't do anything at all, including not allow turning on in nested configs.

In ngx_pagespeed "on" means "on" and "off" means approximately "unplugged".  This is confusing, and means you can't do something like:

```
server {
   pagespeed off;
   location /foo {
      pagespeed on;
   }
}
```

We should make mod_pagespeed and ngx_pagespeed more similar here, probably by making ngx_pagespeed act like mod_pagespeed.

jeffkaufman:
This can't go into a patch release for 1.9, but probably could go into 1.10.

jeffkaufman:
What I'm thinking of doing now:

* split "off" into "unplugged" and "standby"
* "unplugged" means what it does in apache (and what "off" means in nginx)
* "standby" means what "off" does in apache
* deprecate "off"

One hitch is that it looks like ngx_pagespeed currently segfaults if you set "unplugged".  Looking into that.

jeffkaufman:
Splitting the segfault off into https://github.com/pagespeed/ngx_pagespeed/issues/1361

jeffkaufman:
Changes proposed in https://github.com/pagespeed/ngx_pagespeed/issues/968#issuecomment-273825700 are now done in https://github.com/pagespeed/mod_pagespeed/issues/1482 and https://github.com/pagespeed/ngx_pagespeed/issues/1365

PR #1365, opened 2017-01-19
standby: add standby mode

jeffkaufman:
Add standby mode for ngx_pagespeed, equivalent to "off" in mod_pagespeed.

With this change "off" is deprecated, and people should use "unplugged" instead.

See also: https://github.com/pagespeed/mod_pagespeed/pull/1482

Fixes: https://github.com/pagespeed/ngx_pagespeed/issues/968

### DIFF
--- a/src/ngx_rewrite_options.cc
+++ b/src/ngx_rewrite_options.cc
@@ -192,15 +192,18 @@ RewriteOptions::OptionScope NgxRewriteOptions::GetOptionScope(
 
 RewriteOptions::OptionSettingResult NgxRewriteOptions::ParseAndSetOptions0(
     StringPiece directive, GoogleString* msg, MessageHandler* handler) {
-  if (IsDirective(directive, "on")) {
-    set_enabled(RewriteOptions::kEnabledOn);
-  } else if (IsDirective(directive, "off")) {
-    set_enabled(RewriteOptions::kEnabledOff);
-  } else if (IsDirective(directive, "unplugged")) {
-    set_enabled(RewriteOptions::kEnabledUnplugged);
-  } else {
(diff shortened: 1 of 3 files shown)


src/ngx_rewrite_options.cc
```diff
@@ -192,15 +192,18 @@ RewriteOptions::OptionScope NgxRewriteOptions::GetOptionScope(
...
+  EnabledEnum enabled;
+  if (!ParseFromString(directive, &enabled)) {
     return RewriteOptions::kOptionNameUnknown;
   }
+  if (enabled == RewriteOptions::kEnabledOff) {
+    // In ngx_pagespeed, for historical reasons, we treat "off" as "unplugged".
+    // Also, "off" is deprecated and people should be using "standby" or
+    // "unplugged" now depending on which sense they want.  See comment on
+    // RewriteOptions::EnabledEnum.
+    enabled = RewriteOptions::kEnabledUnplugged;
```
morlovich: I am kinda wondering if the mod_pagespeed parser shouldn't do the same and we shouldn't kill the Off enum... Seems a little easier to think of this way, even if it would risk making some of the parsing redundant?

jeffkaufman: That's a good idea!  Let me try that.

jeffkaufman: Getting rid of `EnabledOff` turns out to be pretty hard.  We have a nice option parsing abstraction that isn't compatible with the idea that "off" should map to one value in one server and another value in another server.  It would be possible to extend this, but I don't think it's worth it seeing as once we're done parsing the options everything just asks about `enabled()` and `unplugged()` and doesn't see the enums.

PR #1403, opened 2017-04-08
Update pagespeed_libraries_generator.sh for Ubuntu

demlasjr:
pagespeed_libraries_generator.sh script gives error in Ubuntu

'pagespeed_libraries_generator.sh: 17: pagespeed_libraries_generator.sh: URL+=net/instaweb/genfiles/conf/pagespeed_libraries.conf: not found'

I don't understand why there are 2 lines for the URL when it can be simply switched into one.

### DIFF
--- a/scripts/pagespeed_libraries_generator.sh
+++ b/scripts/pagespeed_libraries_generator.sh
@@ -27,8 +27,9 @@
 # Author: vid@zippykid.com (Vid Luther)
 #         jefftk@google.com (Jeff Kaufman)
 
-URL="https://raw.githubusercontent.com/pagespeed/mod_pagespeed/master/net/instaweb/genfiles/conf/pagespeed_libraries.conf"
-curl -L -s -S "$URL" \
+URL="https://raw.githubusercontent.com/pagespeed/mod_pagespeed/master/"
+URL2="net/instaweb/genfiles/conf/pagespeed_libraries.conf"
+curl -L -s -S "$URL""$URL2" \
     | grep ModPagespeedLibrary \
     | while read library size hash url ; do
   echo "  pagespeed Library $size $hash $url;"
--- a/scripts/pagespeed_libraries_generator.sh
+++ b/scripts/pagespeed_libraries_generator.sh
@@ -27,8 +27,7 @@
 # Author: vid@zippykid.com (Vid Luther)
 #         jefftk@google.com (Jeff Kaufman)
 
-URL="https://github.com/pagespeed/mod_pagespeed/raw/master/"
-URL+="net/instaweb/genfiles/conf/pagespeed_libraries.conf"
+URL="https://raw.githubusercontent.com/pagespeed/mod_pagespeed/master/net/instaweb/genfiles/conf/pagespeed_libraries.conf"
 curl -L -s -S "$URL" \
     | grep ModPagespeedLibrary \
     | while read library size hash url ; do


demlasjr:
I signed it!

oschaaf:
Thanks! Fixing the 404 is great, but the style of the project limits lines to 80 characters, which is why the line was split over two lines.. could you amend that?

scripts/pagespeed_libraries_generator.sh
```diff
@@ -27,9 +27,9 @@
 # Author: vid@zippykid.com (Vid Luther)
 #         jefftk@google.com (Jeff Kaufman)
 
-URL="https://github.com/pagespeed/mod_pagespeed/raw/master/"
-URL+="net/instaweb/genfiles/conf/pagespeed_libraries.conf"
-curl -L -s -S "$URL" \
+URL="https://raw.githubusercontent.com/pagespeed/mod_pagespeed/master/"
+URL2="net/instaweb/genfiles/conf/pagespeed_libraries.conf"
```
oschaaf: Sorry -- introducing `URL2` here is not needed, you can reuse URL (see the original code).
And then below you can go 
```
curl -L -s -S "$URL" \
```

instead of 

```
curl -L -s -S "$URL""$URL2" \
```

demlasjr: That's already made and it's not working, that's why I used URL2. Doing what you say make us returning to the original code which is not working. I'm out of idea here...I just found a easy solution to the problem.

oschaaf: I thought that the change from `https://github.com` to `https://raw.githubusercontent.com` is what fixed the problem? I think that that is all that needs to be changed.

demlasjr:
I already tried that and I still get the same error

> 'pagespeed_libraries_generator.sh: 17: pagespeed_libraries_generator.sh: URL+=net/instaweb/genfiles/conf/pagespeed_libraries.conf: not found'

Replacing **https://github.com** with **https://raw.githubusercontent.com** it's useless since `curl -L` follows redirects itself.

oschaaf:
I just ran `scripts/pagespeed_libraries_generator.sh` on a checkout from master, both on CentOS 6 and Ubuntu 14.04.5 the script works fine for me. So now I am wondering: which shell are you using?

demlasjr:
That's weird. I tested it in Ubuntu 16.04 with the last default shell and
it's not working. I'm not home right now, I'll confirm you later.

oschaaf:
@demlasjr Did you get a chance to take a look at which shell you used when you ran the script and encountered the error?

demlasjr:
@oschaaf Really sorry for the delay.

Here is my shell version:

`# echo $0
-bash

# bash --version 
GNU bash, version 4.3.46(1)-release (x86_64-pc-linux-gnu)


# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.2 LTS
Release:	16.04
Codename:	xenial
`

Thank you.

oschaaf:
So I just spun up a clean ubuntu 16.04 vm on google cloud compute, checked out mps, and ran the script.. which worked fine.  So I still do think it may be something specific to your machine, but I'm out of ideas what it could be.

oschaaf:
@demlasjr Closing this PR, but if you still feel `scripts/pagespeed_libraries_generator.sh` still needs a change please let me know

PR #1460, opened 2017-08-21
Update mod_pagespeed testing dependency

oschaaf:
(mod_pagespeed commit a7c39a1b5a14f1422fef19e8e9dcb11074ba1c65)
This should make CI pass tests

### DIFF
--- a/test/pagespeed_test.conf.template
+++ b/test/pagespeed_test.conf.template
@@ -89,6 +89,16 @@ http {
   @@CONTROLLER@@
 
   server {
+    # This is the default vhost for @@SECONDARY_PORT@@
+    # Your request will land here when you pass in an unknown host.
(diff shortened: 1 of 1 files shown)


oschaaf:
Somehow, with this change, CI still fails on system test 'CSS minify with calc function and value 0.'. Looking into that.

oschaaf:
Updated this to include a change that should make the tests pass

morlovich:
LGTM, but this seems like this will be a headache a lot..

oschaaf:
@morlovich It looks like as of git 1.8.2 it is possible to track branches.
https://stackoverflow.com/questions/9189575/git-submodule-tracking-latest.

Would be nice if we could use that, but I'm not sure if it is a good idea to assume git 1.8.2+ is available to everyone.

Issue #1465, opened 2017-09-02
send_out_headers_and_body() return code is not checked


mxwell:
Return value of the method `send_out_headers_and_body(..)` is not checked, even though it could return an error code:

https://github.com/pagespeed/ngx_pagespeed/blob/65761a2393cc62197129156eb7d2c9374d1aacfe/src/ngx_pagespeed.cc#L2707

I am not sure if it is a real issue, it just looks suspicious to me.

oschaaf:
Good catch. I think that this actually was discussed earlier in a private email thread a long time ago, in the context of a h/2 related bug we were looking in to at the time. The ignores return value was assessed to unrelated at the time, and subsequently forgotten. But I think it would still make sense to properly pass through errors here. I'll take a look at that.

oschaaf:
Fixed in https://github.com/pagespeed/ngx_pagespeed/pull/1468

PR #1468, opened 2017-09-07
Fix ignored return code in ps_simple_handler()

oschaaf:
Fix for https://github.com/pagespeed/ngx_pagespeed/issues/1465


ashishk-1:
Fix looks good to me.

PR #1482, opened 2017-09-25
Proposed changes for releasing 1.12.34.3-stable (ngx_pagespeed only)

oschaaf:

### DIFF
--- a/src/ngx_pagespeed.cc
+++ b/src/ngx_pagespeed.cc
@@ -3007,13 +3007,17 @@ ngx_int_t ps_preaccess_handler(ngx_http_request_t* r) {
   ph = cmcf->phase_engine.handlers;
 
   i = r->phase_handler;
+
   // move handlers before try_files && content phase
+  // As of nginx 1.13.4 we will be right before the try_files module
+  #if (nginx_version < 1013004)
   while (ph[i + 1].checker != ngx_http_core_try_files_phase &&
          ph[i + 1].checker != ngx_http_core_content_phase) {
     ph[i] = ph[i + 1];
     ph[i].next--;
     i++;
   }
+  #endif
 
   // insert ps phase handler
   ph[i].checker = ps_phase_handler;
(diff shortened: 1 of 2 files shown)


oschaaf:
Note: I added a significant change after testing, to amend a mistake I introduced earlier: 3a4fce0ef92cecb4f4a27c1f68306dc20c9f03ed. I'll backport that to master.

oschaaf:
Currently running some more manual tests with this, I plan to merge this once I'm convinced we can release with these changes. Which will probably be somewhere tomorrow. After that we can retag `latest-stable`, and send out the announcement. 

@jmarantz should we also retag `latest-beta` in this case?

jmarantz:
Yes I would not have latest-beta lag latest-stable.

oschaaf:
The output of manually running the system tests is looking good: only the "unauth inlining" tests fail, but that was to be expected because of some recent infrastructural changes.

One issue remaining: I noticed that this emits `X-Page-Speed: 1.12.34.2-0` in the response headers. That makes sense, because that is the PSOL version we are linking in, which is where we get the version from. 

However, this may be slightly confusing, because intuitively one would expect `1.12.34.3` to show up.  I'm wondering: perhaps we should somehow distinguish between versioning of the PageSpeed optimization libraries and its derived ports.. but that probably isn't something that makes sense to do right now.

oschaaf:
@jmarantz see my comment above, do you think we should treat emitting `X-Page-Speed: 1.12.34.2-0`  a blocker? To fix that, we'd probably also have to release a clone of mps 1.12.34.2 as 1.12.34.3 -- with a single line change to update the version header..

oschaaf:
@jmarantz Added a change to ensure we emit the correct version: https://github.com/pagespeed/ngx_pagespeed/pull/1482/commits/04030050f8cfa8ec71b4edee4fbb197682c3187e
LGTY?

PR #1486, opened 2017-10-05
set buildtype from env PSOL_BUILDTYPE.

sawanoboly:
Hi,

I gave the PSOL_BINARY environment variable to skip queries due to # 1332 change on build with `--with-debug`. Then, since `buildtype=Debug` is set, there is no file to include.

I want to use `buildtype=$PSOL_BUILDTYPE` if there is a PSOL_BUILDTYPE environment to avoid that case.

### DIFF
--- a/config
+++ b/config
@@ -19,6 +19,7 @@
 
 mod_pagespeed_dir="${MOD_PAGESPEED_DIR:-unset}"
 position_aux="${POSITION_AUX:-unset}"
+psol_buildtype="${PSOL_BUILDTYPE:-unset}"
 
 if [ "$mod_pagespeed_dir" = "unset" ] ; then
   mod_pagespeed_dir="$ngx_addon_dir/psol/include"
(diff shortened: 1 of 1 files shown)


sawanoboly:
I signed it!

config
```diff
@@ -75,12 +76,17 @@ else
   exit 1
 fi
 
-if [ "$NGX_DEBUG" = "YES" ]; then
-  buildtype=Debug
-  # If we're using a psol tarball that doesn't contain Debug/ (which is the case
-  # from 1.12 onward) then this will be overriden to buildtype=Release below.
+psol_buildtype="${PSOL_BUILDTYPE:-unset}"
```
oschaaf: One request -- can you move this to the top where the others are/declared initialized?
(maybe below `position_aux="${POSITION_AUX:-unset}"`)?

sawanoboly: sure 😃

Issue #1503, opened 2017-11-21
CentOS 6: error: ‘F_SETPIPE_SZ’ was not declared in this scope


centminmod:
Ran into unusual error during Nginx 1.13.7 compile with ngx_pagespeed 1.13.35.1-beta only when attempted on CentOS 6 (with GCC 5.3.1). But no problems with CentOS 7 systems with Nginx 1.13.7

Nginx 1.13.7 was just released http://nginx.org/en/CHANGES

```
                -o objs/addon/src/ngx_caching_headers.o \
                ../ngx_pagespeed-1.13.35.1-beta/src/ngx_caching_headers.cc
[...]

PR #1505, opened 2017-11-23
F_SETPIPE_SZ: check for existence before using

oschaaf:
Kernels before 2.6.35 do not have F_SETPIPE_SZ defined.

Fixes https://github.com/pagespeed/ngx_pagespeed/issues/1503


oschaaf:
@jmarantz unfortunately yes.. maybe we should re-open https://github.com/pagespeed/ngx_pagespeed/issues/1380 for tracking a fix that applies to all systems?

PR #1523, opened 2018-01-08
Unbreak install after repo transfers

oschaaf:
The directory name in the zip we download from github's release
archive will be different now that the project was renamed.
It may be renamed again in the future, so try to obtain the directory
name from the zip.

Fixes https://github.com/apache/incubator-pagespeed-ngx/issues/1521


scripts/build_ngx_pagespeed.sh
```diff
@@ -612,7 +612,8 @@ Not deleting $directory; name is suspiciously short.  Something is wrong."
     nps_downloaded="$TEMPDIR/$nps_downloaded_fname.zip"
     status "Downloading ngx_pagespeed..."
     run wget "$nps_baseurl/$tag_name.zip" -O "$nps_downloaded"
-    nps_module_dir="$BUILDDIR/$nps_downloaded_fname"
+    nps_module_dir=$(unzip -qql "$nps_downloaded" | head -n1 | tr -s ' ' | cut -d' ' -f5-)
```
jmarantz: would be nice to add a comment indicating what this pipe transformation is expected to do.

PR #1529, opened 2018-01-22
Fix build_ngx_pagespeed.sh bug since repo renamed

olala7846:
Since the repo is currently named `/apache/incubator-pagespeed-ngx` instead of `/pagespeed/pagespeed_ngx`, the install script should use `incubator-pagespeed-ngx-${VERSOIN}` instead of `pagespeed_ngx-${VERSION}` as the `nps_downloaded_fname`

### DIFF
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -6,9 +6,9 @@
 # to you under the Apache License, Version 2.0 (the
 # "License"); you may not use this file except in compliance
 # with the License.  You may obtain a copy of the License at
-# 
+#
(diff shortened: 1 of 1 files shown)


oschaaf:
I had hoped that this was already fixed via https://github.com/apache/incubator-pagespeed-ngx/commit/e733a838c35a0f2108768c2d1ead6fb7ee806cf8
Could you paste the exact commands you ran and the error you subsequently ran into?

olala7846:
OMG @oschaaf  you are right, I wan't testing on the latest master... I'm really sorry about that. Thank you so much for the quick response!

PR #1552, opened 2018-04-27
Honor assume yes when installing dependencies

dsanders11:
Currently running `build_ngx_pagespeed.sh` with `--assume-yes` won't automatically send yes when installing dependencies, defeating the purpose. Minor patch which adds `-y` to install commands if `--assume-yes` is used. I may have missed a spot, but I think I got all install commands.

### DIFF
--- a/scripts/build_ngx_pagespeed.sh
+++ b/scripts/build_ngx_pagespeed.sh
@@ -484,10 +484,16 @@ add support for dynamic modules in a way compatible with ngx_pagespeed until
   extra_flags=()
   # Now make sure our dependencies are installed.
   if "$DO_DEPS_CHECK"; then
+    INSTALL_FLAGS=""
+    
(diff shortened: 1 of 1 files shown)


oschaaf:
Code change looks good, and in terms of function it looks like this makes a lot of sense. 
I have to dive in some more to do a proper review, I'll try to do that in the coming days.
One thing I wonder about is that in CI the unattended mode seems to work as-is, 
but that also has `--devel` specified:
https://github.com/apache/incubator-pagespeed-ngx/blob/master/.travis.yml#L23

dsanders11:
Perhaps all the dependencies are installed in the case of `--devel` (or already on the server). I believe if it found all the dependencies it would skip that step so it wouldn't run into the needing input.

I was building on the Docker `ubuntu` image when I ran into this, for reference.

PR #1558, opened 2018-05-13
Fix setting 'gzip_proxied' when applying gzip settings

dsanders11:
This fix should be fairly self-explanatory when viewing the code, but the code was setting the wrong command (and may have been screwing with `gzip_http_version`). This meant that `ngx_pagespeed`'s usage of gzip wouldn't work with CDNs as any fetch with a `Via` header would get the non-gzip version of a file.


oschaaf:
LGTM, nice catch! 
/cc @keesspoelstra

PR #1559, opened 2018-05-13
Properly iterate modules looking for gzip commands

dsanders11:
This change may or may not be needed. I spent a frustratingly long time trying to get gzip working with `ngx_pagespeed` in production, as I kept getting getting the `gzip command not found` error. I eventually got it working, and thought this change was the key, but I tried excluding it and my setup still worked, so now I'm not sure. During the journey of trying to get things working I changed to NGINX 1.14.0 as well as built `ngx_pagespeed` from a git checkout (so I could build up my own fork), so multiple things changed along the way. I unfortunately don't have the time to do a full test of changing a single variable at a time to find what really fixed my issues.

That said, NGINX does [list this change](https://www.nginx.com/blog/nginx-dynamic-modules-how-they-work/#ngxToCycle) as the "proper" way to iterate modules when writing a dynamic module. It's not mentioned, but presumably this doesn't interfere with building as a static module.

@oschaaf, perhaps you can take a look at the change and make a judgement call. I do believe there's a screw loose somewhere with gzip support (I was definitely getting `gzip command not found` with NGINX 1.12.2 and `ngx_pagespeed` v1.13.35.2-stable despite the gzip module being included, confirmed multiple times over), and this may be it. In theory even if this isn't it, this change should be "harmless".

### DIFF
--- a/src/ngx_gzip_setter.cc
+++ b/src/ngx_gzip_setter.cc
@@ -108,10 +108,10 @@ void NgxGZipSetter::Init(ngx_conf_t* cf) {
 #if (NGX_HTTP_GZIP)
   bool gzip_signature_mismatch = false;
   bool other_signature_mismatch = false;
-  for (int m = 0; ngx_modules[m] != NULL; m++) {
-    if (ngx_modules[m]->commands != NULL) {
-      for (int c = 0; ngx_modules[m]->commands[c].name.len; c++) {
-        ngx_command_t* current_command =& ngx_modules[m]->commands[c];
+  for (int m = 0; cf->cycle->modules[m] != NULL; m++) {
+    if (cf->cycle->modules[m]->commands != NULL) {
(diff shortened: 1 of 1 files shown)


dsanders11:
The CI failure seems to be a transient failure relating to the job taking too long. Might have had to do with two PRs being submitted in close succession and the builds overlapping? The build history shows the other PR build took 5 minutes longer than the previous two builds two weeks ago, and then this one took even longer (they overlapped significantly).

oschaaf:
Thanks! I suspect that this is effectively a no-op, but regardless I feel it is a solid improvement as it makes ngx_pagespeed comply with the recommendation from nginx that you pointed out.

@keesspoelstra can you take a look and approve/merge?

oschaaf:
(Also -- CI is passing, I restarted it. The previous try timed out indeed).

Issue #1606, opened 2018-11-23
Build instruction is wrong


dibyendu:
**Similar issue:** #1524 

The instructions provided here: [https://www.modpagespeed.com/doc/build_ngx_pagespeed_from_source](https://www.modpagespeed.com/doc/build_ngx_pagespeed_from_source) is giving wrong url error

#### First download ngx_pagespeed:

```bash
#[check the release notes for the latest version]
NPS_VERSION=1.13.35.1-beta
cd
wget https://github.com/apache/incubator-pagespeed-ngx/archive/v${NPS_VERSION}.zip
unzip v${NPS_VERSION}.zip
nps_dir=$(find . -name "*pagespeed-ngx-${NPS_VERSION}" -type d)
cd "$nps_dir"
NPS_RELEASE_NUMBER=${NPS_VERSION/beta/}
NPS_RELEASE_NUMBER=${NPS_VERSION/stable/}
psol_url=https://dl.google.com/dl/page-speed/psol/${NPS_RELEASE_NUMBER}.tar.gz
[ -e scripts/format_binary_url.sh ] && psol_url=$(scripts/format_binary_url.sh PSOL_BINARY_URL)
wget ${psol_url}  # this line is causing ERROR
tar -xzvf $(basename ${psol_url})  # extracts to psol/
```

#### Output

```
ABC@XYZ:~/incubator-pagespeed-ngx-1.13.35.1-beta$ wget ${psol_url}
--2018-11-23 15:00:20--  https://dl.google.com/dl/page-speed/psol/1.13.35.1-$BIT_SIZE_NAME.tar.gz
Resolving dl.google.com (dl.google.com)... 172.217.12.174
Connecting to dl.google.com (dl.google.com)|172.217.12.174|:443... connected.
HTTP request sent, awaiting response... 404 Not Found
2018-11-23 15:00:20 ERROR 404: Not Found.
```

oschaaf:
closing, resolved via https://github.com/apache/incubator-pagespeed-ngx/pull/1607

PR #1607, opened 2018-11-26
removing the word boundary \b

dibyendu:
Because of the word boundary \b sed fails to replace $BIT_SIZE_NAME from the binary url in macOS.

Related issue: #1606


oschaaf:
LGTM, thanks!

PR #1617, opened 2019-01-13
Update Nginx in Docker

vitaliytv:
Stable from 1.14.1 to 1.14.2
Mainline from 1.15.6 to 1.15.8

### DIFF
--- a/docker/alpine-3.8/nginx-mainline/Dockerfile
+++ b/docker/alpine-3.8/nginx-mainline/Dockerfile
@@ -79,7 +79,7 @@ FROM alpine:$ALPINE_VERSION AS nginx
 ARG NGX_PAGESPEED_TAG=v1.13.35.2-stable
 
 # Check http://nginx.org/en/download.html for the latest version.
-ARG NGINX_VERSION=1.15.6
+ARG NGINX_VERSION=1.15.8
 ARG NGINX_PGPKEY=520A9993A1C052F8
 ARG NGINX_BUILD_CONFIG="\
         --prefix=/etc/nginx \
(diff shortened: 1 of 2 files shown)


nberlee:
Have you tested this?
Everything broke in Alpine Edge because of the LibreSSL to OpenSSL 1.1 switch. These are basically the same changes I did in https://github.com/nberlee/incubator-pagespeed-ngx but as the build failed I wanted to repair it first before upping the version number.

Also in the 1.14.2 releasenote it says that its now compiling with GCC 8.1 and there the patch of nginx is not needed anymore.

vitaliytv:
Sorry, my fault with alpine edge, can I exclude edge?

sachk:
Shouldn't we get ngx_pagspeed working on OpenSSL 1.1 instead of just installing LibreSSL? OpenSSL seems to be much more popular, it's the default on alpine now and it supports TLS 1.3 unlike LibreSSL.

extremeshok:
I have a docker image working and always updated with the latest nginx mainline (1.7.3), openssl 1.1.1x and the latest pagespeed.

https://cloud.docker.com/repository/docker/extremeshok/nginx-pagespeed

zamanuhina:
> 
> 
> I have a docker image working and always updated with the latest nginx mainline (1.7.3), openssl 1.1.1x and the latest pagespeed.
> 
> https://cloud.docker.com/repository/docker/extremeshok/nginx-pagespeed

Hello, backward compatibility?

If I just change pagespeed/nginx-pagespeed to extremeshok/nginx-pagespeed should everything work right away?

oschaaf:
Closing this in an effort to clean up stale pull requests (but feel free to re-open if you want to pick this up again).

PR #1700, opened 2020-07-29
Avoids relying on Bash

wernight:
This script runs fine with just Dash (SH) instead of Bash. Alpine Linux and others don't have Bash.


oschaaf:
Lgtm — thanks

PR #1750, opened 2022-06-27
Fixes for NGINX nginx 1.23.0

oschaaf:
As of nginx 1.23 the cache control headers are re-implemented as al linked list.
This reworks our code that manipulates this header to work with that.

Still a draft: this change probably needs to be conditioned against nginx version
with ifdefs, so older version can still be built as well. Hence, still a draft.

Attempts to fix https://github.com/apache/incubator-pagespeed-ngx/issues/1749


oschaaf:
closing this in favor of https://github.com/apache/incubator-pagespeed-ngx/pull/1751

